PR #31, opened 2015-10-02
Detect service from url

pape-io:
When using `includes` to include linked objects use known services for doing the request to be able to apply endpoint configurations from their service definitions (e.g. Auth).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,3 +9,5 @@ gemspec
 # your gemspec. These might include edge Rails or gems from your path or
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
+
+gem 'lhc', '~> 1.2.0', path: '../lhc'
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source 'https://rubygems.intra.local.ch/'
+source 'https://rubygems.org/'
 
 # Declare your gem's dependencies in lhs.gemspec.
 # Bundler will treat runtime dependencies like base dependencies, and
--- a/Gemfile
+++ b/Gemfile
@@ -9,5 +9,3 @@ gemspec
 # your gemspec. These might include edge Rails or gems from your path or
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
-
-gem 'lhc', '~> 1.2.0', path: '../lhc'
--- a/lib/lhs.rb
+++ b/lib/lhs.rb
@@ -4,3 +4,10 @@ module LHS
 end
 
 Gem.find_files('lhs/**/*.rb').each { |path| require path }
+
+# Preload all the services that are defined in app/services
+class Engine < Rails::Engine
+  initializer 'Load all services' do |app|
+    Dir.glob(app.root.join('app/services/**.rb')).each {|file| require file }
+  end
+end
--- a/lib/lhs.rb
+++ b/lib/lhs.rb
@@ -8,6 +8,6 @@ Gem.find_files('lhs/**/*.rb').each { |path| require path }
 # Preload all the services that are defined in app/services
 class Engine < Rails::Engine
   initializer 'Load all services' do |app|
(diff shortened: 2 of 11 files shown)


Gemfile
```diff
@@ -9,3 +9,5 @@ gemspec
 # your gemspec. These might include edge Rails or gems from your path or
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
+
+gem 'lhc', '~> 1.2.0', path: '../lhc'
```
skaestle: local path is referenced

lib/lhs.rb
```diff
@@ -4,3 +4,10 @@ module LHS
 end
 
 Gem.find_files('lhs/**/*.rb').each { |path| require path }
+
+# Preload all the services that are defined in app/services
+class Engine < Rails::Engine
+  initializer 'Load all services' do |app|
+    Dir.glob(app.root.join('app/services/**.rb')).each {|file| require file }
```
skaestle: maybe you wanna include sub directories?

lib/lhs/concerns/service/endpoints.rb
```diff
@@ -18,6 +19,15 @@ def endpoint(url, options = nil)
         endpoint = LHC::Endpoint.new(url, options)
         instance.sanity_check(endpoint)
         instance.endpoints.push(endpoint)
+        LHS::Service::Endpoints.all ||= {}
```
skaestle: can you initialize it in the initializer method?

pape-io: initalizer method for a module?

lib/lhs/concerns/service/request.rb
```diff
@@ -15,38 +15,34 @@ def request(options)
...
-    end
-
-    def single_request(options)
-      response = LHC.request(process_options(options))
-      data = LHS::Data.new(response.body, nil, self.class, response.request)
-      handle_includes(data) if includes
-      data
+    # Convert URLs in options to endpoint templates
+    def convert_options_to_endpoints(options)
+      if options.is_a? Array
```
skaestle: i would do: `options.respond_to?(:map)`

lib/lhs/concerns/service/endpoints.rb
skaestle: agreed to leave it as it is

lib/lhs/concerns/service/request.rb
```diff
@@ -59,36 +55,77 @@ def handle_includes(data)
...
+      params.delete(:params)
+      params.merge!(explicit_params) if explicit_params
+    end
+
+    def multiple_requests(options)
+      options = options.map { |options| process_options(options) }
+      responses = LHC.request(options)
+      data = responses.map{ |response| LHS::Data.new(response.body, nil, self.class, response.request) }
+      data = LHS::Data.new(data, nil, self.class)
+      handle_includes(data) if includes
```
skaestle: `self.` should be superfluous

lib/lhs/concerns/service/request.rb
```diff
@@ -59,36 +55,77 @@ def handle_includes(data)
...
+      options = (endpoint.options || {}).merge(options)
+      options[:url] = compute_url!(options[:params]) unless options.key?(:url)
+      merge_explicit_params!(options[:params])
+      options.delete(:params) if options[:params] && options[:params].empty?
+      options
+    end
+
+    def service_for_options(options)
+      services = []
+      if options.is_a?(Array)
```
skaestle: again, imo: `options.respond_to?(:each)`

skaestle:
documentation needs update, but other than that: :+1:

lib/lhs/concerns/service/request.rb
pape-io: http://ruby-doc.org/core-2.2.0/Class.html 
69ee5fbbec1cc6d10a91fbdb34ac5503

Issue #33, opened 2015-10-06
double id issues


pape-io:
```
endpoint ':datastore/v2/places', auth: { bearer: -> { RequestStore.store[:bearer_token] } }
endpoint ':datastore/v2/places/:id', auth: { bearer: -> { RequestStore.store[:bearer_token] } }
```

```
ETHON: performed EASY effective_url=HTTP://local.ch:8082/v2/places/ZW9OJyrbt4OZE9ueu80w-A/ZW9OJyrbt4OZE9ueu80w-A response_code=404 return_code=ok total_time=0.031823
LHC::NotFound: get local.ch:8082/v2/places/ZW9OJyrbt4OZE9ueu80w-A/ZW9OJyrbt4OZE9ueu80w-A
Params: {:auth=>{:bearer=>#<Proc:0x007fb7b3fe9a08@/Users/Sebastian/Work/Extern/local/location-app/app/services/place.rb:4 (lambda)>}, :url=>"local.ch:8082/v2/places/ZW9OJyrbt4OZE9ueu80w-A/ZW9OJyrbt4OZE9ueu80w-A"}
Response Code: 404

```

pape-io:
Fixed with https://github.com/local-ch/lhs/releases/tag/v1.0.0

PR #36, opened 2015-10-07
No implicit id parameter handling anymore

pape-io:
It was a bad idea to handle id as an implicit parameter.

It led to an issue where the id was passed as url parameter twice: https://github.com/local-ch/lhs/issues/33

So I decided to completely remove the implicit id handling and make it transparent.

## Major version increase

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -18,7 +18,9 @@ You can also add request options for an endpoint (see following example).
 class Feedback < LHS::Service
 
   endpoint ':datastore/v2/content-ads/:campaign_id/feedbacks'
+  endpoint ':datastore/v2/content-ads/:campaign_id/feedbacks/:id'
   endpoint ':datastore/v2/feedbacks', cache: true, cache_expires_in: 1.day
+  endpoint ':datastore/v2/feedbacks/:id', cache: true, cache_expires_in: 1.day
 
 end
 ```
--- a/lib/lhs/concerns/service/endpoints.rb
+++ b/lib/lhs/concerns/service/endpoints.rb
@@ -57,7 +57,6 @@ class LHS::Service
     def compute_url!(params)
       endpoint = find_endpoint(params)
       url = endpoint.compile(params)
-      url +=  "/#{params.delete(:id)}" if params && params[:id]
       endpoint.remove_interpolated_params!(params)
       url
     end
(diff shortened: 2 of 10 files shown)


kot-begemot:
Did you cover that issue with test, to be sure that you actually fixed it?

pape-io:
That one is removing a feature. Would be new to me, that thats testable?

docs/services.md
```diff
@@ -18,7 +18,9 @@ You can also add request options for an endpoint (see following example).
 class Feedback < LHS::Service
 
   endpoint ':datastore/v2/content-ads/:campaign_id/feedbacks'
+  endpoint ':datastore/v2/content-ads/:campaign_id/feedbacks/:id'
   endpoint ':datastore/v2/feedbacks', cache: true, cache_expires_in: 1.day
+  endpoint ':datastore/v2/feedbacks/:id', cache: true, cache_expires_in: 1.day
```
eduardschaeli: Why set "cache: true" when setting a chace expiration kinda implies that you're using a cache? Couldn't you get rid of the "cache: true" param?

kot-begemot:
Well you solved an issue, isn't it?

docs/services.md
Averethel: Valid point but out of scope of this PR. Can you create an issue?

Averethel:
@pape-io :+1:  for implementation and :+1: for being explicit. ILIKE!

docs/services.md
pape-io: Looks like it would be worth improving.
Moved the issue here: https://github.com/local-ch/lhc-core-interceptors/issues/13

Averethel: :+1:

eduardschaeli: Ui, you're too quick for me, was still reading the PR ;)

eduardschaeli:
Cool as far as I understand this project :+1:

pape-io:
There is your test https://github.com/local-ch/lhs/commit/8f36a3b2d4160e8a81da13e1cb3858af7cc7e3fa
@kot-begemot

kot-begemot:
http://24.media.tumblr.com/5bbd35e859ecfc817dbbf592bf6ea3ec/tumblr_mukjvk8egE1swuewlo1_250.gif

PR #39, opened 2015-10-07
Mappings for nested data

pape-io:
Allows mappings based on Service definitions also for nested data.


Averethel:
:+1: there's a bit magic there but I think it's a good magic ;)

skaestle:
:+1:

kot-begemot:
:+1:

PR #41, opened 2015-10-07
~ improve all to also work with responses without limit

pape-io:


acoiro:
code looks good :+1: , but I'm having hard time understanding why you use `data.count` as limit

kot-begemot:
:+1: good branch name.

acoiro:
@kot-begemot "all" doesn't mean "everything", it means the method `all`

Issue #43, opened 2015-10-09
load! is not merging data properly


pape-io:
load! is loading a link, but does not merge the data properly after fetching it.
So the data is not accessible.

PR #44, opened 2015-10-09
! fix load! merging data

pape-io:
Fix https://github.com/local-ch/lhs/issues/43

### DIFF
--- a/spec/proxy/load_spec.rb
+++ b/spec/proxy/load_spec.rb
@@ -32,14 +32,14 @@ describe LHS::Proxy do
   context 'load' do
 
     it 'is loading data remotely when not present yet' do
-      link.load!.id
+      expect(link.load!.id).to be
+      expect(link.id).to be
     end
 
     it 'can be reloaded' do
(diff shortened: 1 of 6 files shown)


kot-begemot:
:+1: looks good

spec/proxy/load_spec.rb
```diff
@@ -32,14 +32,14 @@ class SomeService < LHS::Service
   context 'load' do
 
     it 'is loading data remotely when not present yet' do
-      link.load!.id
+      expect(link.load!.id).to be
+      expect(link.id).to be
```
acoiro: this means present?

acoiro:
:+1:

spec/proxy/load_spec.rb
pape-io:
```
 expect(obj).to be         # passes if obj is truthy (not nil or false)
```
https://www.relishapp.com/rspec/rspec-expectations/v/3-3/docs/built-in-matchers/be-matchers

Issue #30, opened 2015-09-29
LHS::Collection: present? and empty? are not forwarded to items


pape-io:
LHS::Collection does not forward present? and empty? correctly to the items. 
Therefor `present?` is true as soon as an object is around but items is empty
and `empty?` fails as not being implemented.

PR #47, opened 2015-10-26
! fix destroy with includes and delegates for item and collection

pape-io:
- Fixes https://github.com/local-ch/lhs/issues/30
- Fixes destroy in combination with includes

### DIFF
--- a/lib/lhs/collection.rb
+++ b/lib/lhs/collection.rb
@@ -75,5 +75,8 @@ class LHS::Collection < LHS::Proxy
     delegate :last, to: :raw
     delegate :sample, to: :raw
     delegate :[], to: :raw
+    delegate :present?, to: :_raw
+    delegate :blank?, to: :raw
+    delegate :empty?, to: :raw
   end
 end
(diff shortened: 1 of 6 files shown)


lib/lhs/collection.rb
```diff
@@ -75,5 +75,8 @@ def each(&block)
     delegate :last, to: :raw
     delegate :sample, to: :raw
     delegate :[], to: :raw
+    delegate :present?, to: :_raw
```
kot-begemot: I'd say that it is better to list all those methods in one line, since delegate method accepts multiple multiple parameters

http://apidock.com/rails/Module/delegate

or group it for better readability

https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/activerecord/lib/active_record/querying.rb

lib/lhs/item.rb
```diff
@@ -8,6 +8,10 @@ class LHS::Item < LHS::Proxy
   include Save
   include Update
 
+  delegate :present?, to: :_raw
```
kot-begemot: Same as previous comment

kot-begemot:
A small suggestion. Otherwise looks fine :+1:

Averethel:
:+1:

skaestle:
:+1:

Issue #48, opened 2015-10-29
Errors are not mapped correctly anymore


batiste:
LHS expect "fields"

PR https://github.com/local-ch/lhs/pull/50

https://github.com/local-ch/lhs/blob/master/lib/lhs/errors.rb#L72

But "field_errors" is returned by the backend TTP://datastore-stg.lb-service.sunrise.intra.local.ch/v2/localch-accounts

    => #<LHS::Errors:0x007fc3face6f10 @message="Some data in the request body failed validation. Inspect the field errors for details.", @messages={}>
    [4] pry(#<RegisterController>)> JSON.parse(@user._parent._request.raw.response.options[:response_body])
    => {"status"=>400,
     "message"=>"Some data in the request body failed validation. Inspect the field errors for details.",
     "field_errors"=>
      [{"code"=>"REQUIRED_PROPERTY_VALUE", "path"=>["name"], "message"=>"The property value is required; it cannot be null, empty, or blank"},
       {"code"=>"INVALID_PROPERTY_VALUE_FORMAT", "path"=>["email"], "message"=>"The property value format is invalid. For example, specifying the “12/15/2012” date format when “2012-12-15” is expected."},
       {"code"=>"REQUIRED_PROPERTY_VALUE", "path"=>["name"], "message"=>"The property value is required; it cannot be null, empty, or blank"},
       {"code"=>"REQUIRED_PROPERTY_VALUE", "path"=>["phone"], "message"=>"The property value is required; it cannot be null, empty, or blank"}]}

batiste:
Can be tested here: https://github.com/local-ch/ws-customer-center/pull/192

PR #50, opened 2015-11-20
Handle 2 format of error messages

batiste:
https://github.com/local-ch/lhs/issues/48

### DIFF
--- a/lib/lhs/errors.rb
+++ b/lib/lhs/errors.rb
@@ -64,7 +64,7 @@ class LHS::Errors
 
   private
 
-  def parse_old_errors(errors)
+  def parse_fields(errors)
     parsed = {}
     errors.each do |field|
       name = field['name'].to_sym
@@ -76,7 +76,7 @@ class LHS::Errors
     parsed
   end
 
-  def parse_errors(errors)
+  def parse_field_errors(errors)
     parsed = {}
     errors.each do |field|
       name = field['path'].first.to_sym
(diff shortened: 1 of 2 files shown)


lib/lhs/errors.rb
```diff
@@ -64,19 +64,38 @@ def empty?
...
-    json = JSON.parse(response.body)
-    return messages unless json['fields']
-    json['fields'].each do |field|
+  def parse_old_errors(errors)
+    parsed = {}
+    errors.each do |field|
       name = field['name'].to_sym
-      messages[name] ||= []
+      parsed[name] ||= []
       field['details'].each do |detail|
```
pape-io: That was related to active records related internal error messages... 
Renaming looks arbitrary on first sight

lib/lhs/errors.rb
```diff
@@ -64,19 +64,38 @@ def empty?
...
+      parsed[name].push(field['code'])
+    end
+    parsed
+  end
+
+  def messages_from_response(response)
+    return {} if !response.body.is_a?(String) || response.body.length.zero?
+    json = JSON.parse(response.body)
+    if json['fields']
+      return parse_old_errors(json['fields'])
```
pape-io: Should not be old & new as this is just the case for local
'''
%w(fields field_errors).each do |key|
  parse_errors(body, key)
end
'''

I suggest merging thepossible errors

pape-io: Then you would also bot end up having two methods for parsing

pape-io:
Would not call it old and new, as this is just the case for local and the error key will might be dynamic in the future.

Also please try to have just one parse method that you provide the specific error key.

Otherwise :+1:

pape-io:
:+1:

PR #55, opened 2015-11-27
+ LHS::Item#valid?

pape-io:


lib/lhs/concerns/item/validation.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  module Validation
+    extend ActiveSupport::Concern
+
+    def valid?
+      self.errors = nil
+      fail 'No validation endpoint found!' unless validation_endpoint
+      service = LHS::Service.for_url(validation_endpoint.url)
+      begin
+        !! service.request(
```
Averethel: wouldn't simple returning `true` be more readable?

Averethel:
:+1:

lib/lhs/concerns/item/validation.rb
pape-io: `true` ;)

PR #59, opened 2015-12-17
! consistent internal data

pape-io:
Improve internal data structure by sticking to symbols always!


acoiro:
:+1: when tests are green

PR #60, opened 2015-12-21
! internal data structure in case of building

pape-io:
Handle the ActionController::Parameters case as its not really a Hash.


acoiro:
:+1:

skaestle:
:+1:

PR #65, opened 2015-12-29
Correctly handle recursion

acoiro:


pape-io:
To understand recursion...
you have to understand recursion.

pape-io:
:+1:

Issue #56, opened 2015-11-27
Service new == build


pape-io:
Currently `Service#new` creates a new instance of a LHS::Service.

As ActiveRecords behaviour for `new` is to create and return a new instance of this business object and not an mapper between object and db column, we should reconsider if we want to also return a freshly build business object that you can work with and persist with `save`.

ActiveRecord unified the behaviour of `new` and `build` some years ago. They kept `build` as alias for new.

skaestle:
sounds good. however `Service#new` is kinda weird. it makes sense on activerecord because you send it to the model. what would be the equivalent in LHS terms? I guess `Service`...

Averethel:
I agree, `new` seems to be a bit awkward name, `build` gives better overview of what's going on.

acoiro:
`new` is fine, the `Service` name is actually a bit weird since it's more close to a model/mapper.

We want to use the `User` LHS service with RailsAuth which defines the `current_user` method `current_user ||= RailsAuth.user_class.new`

For this reason, and also because we want to have an interface similar to ActiveRecord, I would alias `new` to `build`.

pape-io:
@skaestle: Actually `new` on ActiveRecord creates a new (OR)Mapper that persists itself on the DB later on with `save`.

I agree with @acoiro that calling it `Service` is maybe the problem. 

We brainstormed a little bit and came up with something like`RESTRecord` to replace the term `Service` in LHS. To be discussed.

pape-io:
added `new` as alias for `build` in https://github.com/local-ch/lhs/pull/66 mainly for https://github.com/local-ch/lhs/issues/56#issuecomment-167957993

pape-io:
https://github.com/local-ch/lhs/releases/tag/v2.1.0

pape-io:
https://github.com/local-ch/lhs/releases/tag/v3.0.0

PR #66, opened 2015-12-30
Initalize with new

pape-io:
https://github.com/local-ch/lhs/issues/56


acoiro:
Please add it to docs

PR #68, opened 2016-01-28
+ ciderized

Averethel:
rubocop and rspec on cider


spec/record/includes_spec.rb
```diff
@@ -161,6 +154,7 @@ class SomeInterceptor < LHC::Interceptor; end
       end
 
       it 'uses interceptors for included links from known services' do
+        # rubocop:disable RSpec/InstanceVariable
```
pape-io: Let's try to change the test, so that we do not have to disable rubocop here.

Issue #70, opened 2016-02-10
automatic date parsing fails for date_time


ryabrody:
```bash
@_raw=
  {:no_advertising=>true,
   :answered=>true,
   :customer=>false,
   :reasons=>["SELLING", "BAD_TIME", "UNFRIENDLY"],
   :additional_reason=>"Lorem Ipsum Dolor Sit amet.",
   :last_name=>"Müller",
   :first_name=>"Hans",
   :number=>"41123456789",
   :telemarketer_number=>"41586354324",
   :occasion_date_time=>"2016-07-09T13:45:00+00:00"},
 @_request=nil,
 @_service=TelemarketingComplaint>
[4] pry(TelemarketingComplaint)> a.occasion_date_time
=> Sat, 09 Jul 2016
[5] pry(TelemarketingComplaint)> a.occasion_date_time
=> Sat, 09 Jul 2016
[6] pry(TelemarketingComplaint)> a.occasion_date_time.class
=> Date
```

pape-io:
Backport to v2

pape-io:
FIX for v3 https://github.com/local-ch/lhs/pull/82

pape-io:
https://github.com/local-ch/lhs/releases/tag/v3.0.5

PR #82, opened 2016-02-23
! different formats for automated date time parsing

pape-io:
Fixes https://github.com/local-ch/lhs/issues/70


Averethel:
:+1:

batiste:
Coudn't this be done by some kind of library?

pape-io:
Don't want to add blown up libraries, and don't want to spent hours for research right now. If you have one that is not blown up, and it doesn't take you hours to find it, suggest one @batiste

PR #85, opened 2016-02-25
! fix casting items in collections

pape-io:


lib/lhs/collection.rb
```diff
@@ -64,11 +59,27 @@ def initialize(raw, parent, record)
...
         if item.is_a? Hash
-          yield LHS::Data.new(item, @parent, @record)
+          yield cast_item(item)
         else
           yield item
         end
       end
     end
+
+    private
```
ryabrody: check the `private`

acoiro: duplication

lib/lhs/collection.rb
```diff
@@ -64,11 +59,27 @@ def initialize(raw, parent, record)
...
+
+    def cast_item(item)
+      data = LHS::Data.new(item, @parent, @record)
+      return @record.new(data) if @record
```
acoiro: You don't like that?
`@record ? @record.new(data) : data`

acoiro:
Small comment, otherwise :+1:

lib/lhs/collection.rb
pape-io: It's an inlined, private class.... what ever, it looks weird, will try to refactor it

lib/lhs/concerns/collection/internal_collection.rb
```diff
@@ -0,0 +1,41 @@
+require 'active_support'
+
+class LHS::Collection < LHS::Proxy
```
ryabrody: I do not get why you are defining this class twice

pape-io: It's a Concern.

PR #86, opened 2016-02-25
+ Use defined setters in initalizer

### LINKED ISSUE #84: Custom setter are not called when using `#new`
# Example

## Definition
```ruby
class Feedback < LHS::Record
  def ratings=(value)
    _raw[:ratings] = rating_to_raw(value)
  end

  def rating_to_raw(value)
    return [] if value.nil?

    value.map do |k, v|
      {
        dimension: k.to_s,
        rating: v
      }
    end
  end
end
```

## Method call
```ruby
Feedback.new(ratings: { quality: 3 })
```

## Expected result
```ruby
<Feedback:... @data={:ratings=>[{ dimension: "quality", rating: 3 }]}>
```

## Received result
```ruby
<Feedback:... @data={:ratings=>{ quality: 3 }}>
```

## Suggested solution
In the initializer iterate over keys, if there's custom setter for the key use it otherwise assign to `_raw[<key>]`

Averethel:
Fixes #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Uses the `:datastore/v2/content-ads/:campaign_id/feedbacks` endpoint.
 If no record is found an error is raised.
 
 ## Proxy
-Instead of mapping data when it arrives from the backend, the proxy makes data accessible when you access it, not when you fetch it. The proxy is used to access data and it is divided in `Collection` and `Item`. 
+Instead of mapping data when it arrives from the backend, the proxy makes data accessible when you access it, not when you fetch it. The proxy is used to access data and it is divided in `Collection` and `Item`.
 
 `find` can also be used to find a single uniqe record with parameters:
 
@@ -119,7 +119,7 @@ data.total # 998
 ```ruby
 Feedback.find_each(start: 50, batch_size: 20, params: { has_reviews: true }) do |feedback|
   # Iterates over each record. Starts with record nr. 50 and fetches 20 records each batch.
-  feedback 
+  feedback
   break if feedback.some_attribute == some_value
 end
 ```
@@ -223,10 +223,10 @@ The implementation is heavily influenced by [http://guides.rubyonrails.org/activ
 ```ruby
   # list of includes
   claims = Claims.includes(:localch_account, :entry).where(place_id: 'huU90mB_6vAfUdVz_uDoyA')
-  
+
   # array of includes
   claims = Claims.includes([:localch_account, :entry]).where(place_id: 'huU90mB_6vAfUdVz_uDoyA')
-  
+
   # Two-level with array of includes
   feedbacks = Feedback.includes(campaign: [:entry, :user]).where(has_reviews: true)
 ```
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ If no record is found, `nil` is returned.
 
 ## Batch processing
 
-**Be carefull using methods for batch processing. They could result in a lot of HTTP requests!**
+**Be careful using methods for batch processing. They could result in a lot of HTTP requests!**
 
 `all` fetches all records from the backend by doing multiple requests if necessary.
 
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Build and persist new items from scratch are done either with `new` or it's alia
 
 ## Custom setters and getters
 
-Sometimes it is the case that you want to have your custom getters and setters and convert the data to backend processable format behind the scenes. It is now supported by the initializer.
+Sometimes it is the case that you want to have your custom getters and setters and convert the data to backend processable format behind the scenes. The initializer will now use custom setter if one is defined
 
 ```ruby
 class Feedback < LHS::Record
--- a/README.md
+++ b/README.md
@@ -182,7 +182,9 @@ If you have an accompanying getter the whole data manipulation would be internal
 
 ```ruby
 class Feedback < LHS::Record
-  ...
+  def ratings=(ratings)
+    _raw[:ratings] = ratings.map { |k, v| { name: k, value: v } }
+  end
 
   def ratings
     Hash[_raw[:ratings].map { |r| [r[:name], r[:value]] }]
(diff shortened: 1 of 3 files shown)


lib/lhs/record.rb
```diff
@@ -29,7 +29,8 @@ def initialize(data = nil)
       else
         data._raw
       end
-    instance_variable_set('@data', instance_data)
```
pape-io: That was used to make raw data visible while debugging on the terminal, now its empty:
```
feedback
=> #<Feedback:0x007fd8611f8308 @data={}>
[2] pry(#<RSpec::ExampleGroups::LHSRecord::New::CustomSetters>)> feedback._raw
=> {:ratings=>[1, 2]}
```

Averethel: So this should have the same content as raw? Okay I'll change it

Averethel: Fixed in 295cb91

README.md
```diff
@@ -162,6 +162,38 @@ Build and persist new items from scratch are done either with `new` or it's alia
...
+class Feedback < LHS::Record
+  ...
+
+  def ratings
+    Hash[_raw[:ratings].map { |r| [r[:name], r[:value]] }]
+  end
+end
+
+feedback = Feedback.new(ratings: { quality: 3 }) # #<Feedback:0x007fc8f9a43598 @data={:ratings=>{:quality=>3}}>
+feedback.ratings # {:quality=>3}
```
pape-io: Shouldn't this be `['quality', 3]` ???

README.md
```diff
@@ -162,6 +162,38 @@ Build and persist new items from scratch are done either with `new` or it's alia
   feedback.save
 ```
 
+## Custom setters and getters
+
+Sometimes it is the case that you want to have your custom getters and setters and convert the data to backend processable format behind the scenes. It is now supported by the initializer.
```
pape-io: I thought, it is also enables you to use the custom setters. It sounds a little bit like, there are custom setters, but you can *only* use them by using the initialiser. Could you add an example for using the  custom setter: `feedback.ratings = whatever`

lib/lhs/record.rb
```diff
@@ -49,4 +50,22 @@ def method_missing(name, *args, &block)
   def respond_to_missing?(name, include_all = false)
     _data.respond_to_missing?(name, include_all)
   end
+
+  private
+
+  def data=(instance_data)
+    return instance_variable_set('@data', instance_data) if !instance_data.is_a?(Hash)
```
pape-io: To be honest, I don't quite get this case.

When instance_data is not a hash, don't change any raw, but update `@data` which is just used for debugging reasons??

lib/lhs/record.rb
```diff
@@ -29,7 +29,8 @@ def initialize(data = nil)
       else
         data._raw
       end
-    instance_variable_set('@data', instance_data)
+
+    send(:data=, instance_data)
```
pape-io: Couldn't we use
```
data = instance_data
```
Would be the same, no?

spec/record/new_spec.rb
```diff
@@ -22,7 +23,45 @@ class Feedback < LHS::Record
...
+            def ratings
+              Hash[_raw[:ratings].map { |r| [r[:name], r[:value]] }]
+            end
+          end
+        end
+
+        it 'uses custom getters to show data for exploration' do
+          feedback = Feedback.new(ratings: { a: 1, b: 2 })
+          expect(feedback.instance_values['data'][:ratings]).to eq(a: 1, b: 2)
+          expect(feedback.ratings).to eq(a: 1, b: 2)
```
pape-io: `uses custom getters to show data for exploration`

This should be: 
```
eq [['a', 1], ['b', 2]]
```

spec/record/new_spec.rb
```diff
@@ -22,7 +23,45 @@ class Feedback < LHS::Record
...
+          def ratings=(ratings)
+            _raw[:ratings] = ratings.map { |k, v| { name: k, value: v } }
+          end
+        end
+      end
+
+      it 'are used by initializer' do
+        feedback = Feedback.new(ratings: { a: 1, b: 2 })
+        expect(feedback.instance_values['data'][:ratings]).to eq([{ name: :a, value: 1 }, { name: :b, value: 2 }])
+        expect(feedback.ratings.raw).to eq([{ name: :a, value: 1 }, { name: :b, value: 2 }])
```
pape-io: Can you check why `feedback.ratings.raw` even works?!?!?!
It shouldn't because it's defined as `_raw` everywhere.

spec/record/new_spec.rb
```diff
@@ -22,7 +23,52 @@ class Feedback < LHS::Record
...
+      end
+
+      it 'can be used directly to change raw data' do
+        feedback = Feedback.new(ratings: { a: 1 })
+        feedback.ratings = { z: 3 }
+        expect(feedback.ratings.first.name).to eq :z
+        expect(feedback.instance_values['data'][:ratings].first[:name]).to eq :z
```
pape-io: When using custom setters for changing data, not initalising, `@data` (the one for debugging) is not yet getting updated.

README.md
```diff
@@ -162,6 +162,38 @@ Build and persist new items from scratch are done either with `new` or it's alia
...
+feedback.ratings # #<LHS::Data:0x007fc8fa6d4050 ... @_raw=[{:name=>:quality, :value=>3}]>
+
+```
+
+If you have an accompanying getter the whole data manipulation would be internal only.
+
+```ruby
+class Feedback < LHS::Record
+  ...
+
```
pape-io: Add comment here that the custom setter is used.

lib/lhs/record.rb
pape-io: `self.data=` would use public `send`.

lib/lhs/record.rb
pape-io: It's guard clause that protects iterating hashes. Like arrays.

spec/record/new_spec.rb
pape-io: Please use `_raw`

Averethel:
@pape-io implemented your suggestion, new inspect LGTM.

How 'bout we merge it?

pape-io:
:+1:

PR #88, opened 2016-03-01
Use new in create

Averethel:
We discovered that `create` does not use new and therefore custom setters are not used.
It is logical to use the initialiser in `create` and build it up from the blocks we already implemented elsewhere.

I also discovered that if url is computed in `save!` the parameter used to compute is not removed from request body, which should happen - it's fixed now.


lib/lhs/concerns/record/create.rb
```diff
@@ -7,19 +7,15 @@ module Create
...
     module ClassMethods
       def create(data = {})
-        create!(data)
-      rescue LHC::Error => e
-        json = JSON.parse(data.to_json)
-        data = LHS::Data.new(json, nil, self, e.response.request)
-        item = LHS::Item.new(data)
-        item.errors = LHS::Errors.new(e.response)
-        data._record_class.new(LHS::Data.new(item, data))
+        object = new(data)
```
pape-io: we should call it record

lib/lhs/concerns/record/create.rb
```diff
@@ -7,19 +7,15 @@ module Create
...
+        object.save
+        object
       end
 
       def create!(data = {})
-        url = compute_url!(data)
-        data = request(url: url, method: :post, body: data.to_json, headers: { 'Content-Type' => 'application/json' })
-        data._record_class.new(data)
+        object = new(data)
```
pape-io: should be called `record`

pape-io:
Small naming issues. The rest is :+1:.

Averethel:
@pape-io Are we good to go on this one?

PR #89, opened 2016-03-02
reproduce errors in specs

ryabrody:
# Error case
It is not possible to spec requests like this when the method `all` is used inside controller:
### controller
```ruby
def index
    @callcenters = Spamnumber.all(type: 'Verified', order_by: '-user_frequency', swiss_number: true)
end
```
### spec
```ruby
let(:callcenters) { [callcenter] }
before :each doe
      stub_request(:get, %r{/spamnumbers}).to_return(body: callcenters.to_json)
end
```
Instead requests have to be stubbed like this:
### spec
```ruby
let(:callcenters) { { items: [callcenter], total: 1 } }
before :each doe
      stub_request(:get, %r{/spamnumbers}).to_return(body: callcenters.to_json)
end
```
This error case is reproduced in the red specs from following spec-file:
```bash
rspec spec/record/all_spec.rb
```

### DIFF
--- a/spec/record/all_spec.rb
+++ b/spec/record/all_spec.rb
@@ -48,5 +48,21 @@ describe LHS::Collection do
       expect(all._proxy).to be_kind_of LHS::Collection
       expect(all.count).to eq 0
     end
+
+    it 'alsow works when there is no total in the stubbing' do
+      stub_request(:get, %r{/feedbacks}).to_return(body: { items: (1..100).to_a }.to_json)
+      all = Record.all
+      expect(all).to be_kind_of Record
+      expect(all._proxy).to be_kind_of LHS::Collection
+      expect(all.count).to eq 0
+    end
+
+    it 'alsow works when there is no key "items" in the stubbing' do
+      stub_request(:get, %r{/feedbacks}).to_return(body: (1..100).to_a.to_json)
+      all = Record.all
+      expect(all).to be_kind_of Record
+      expect(all._proxy).to be_kind_of LHS::Collection
+      expect(all.count).to eq 0
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/record/all_spec.rb
```diff
@@ -48,5 +48,21 @@ class Record < LHS::Record
...
+      expect(all._proxy).to be_kind_of LHS::Collection
+      expect(all.count).to eq 0
+    end
+
+    it 'alsow works when there is no key "items" in the stubbing' do
+      stub_request(:get, %r{/feedbacks}).to_return(body: (1..100).to_a.to_json)
+      all = Record.all
+      expect(all).to be_kind_of Record
+      expect(all._proxy).to be_kind_of LHS::Collection
+      expect(all.count).to eq 0
```
pape-io: Why 0? Shouldn't it be 100, as there are 100 items in the body?

spec/record/all_spec.rb
```diff
@@ -48,5 +48,21 @@ class Record < LHS::Record
       expect(all._proxy).to be_kind_of LHS::Collection
       expect(all.count).to eq 0
     end
+
+    it 'alsow works when there is no total in the stubbing' do
+      stub_request(:get, %r{/feedbacks}).to_return(body: { items: (1..100).to_a }.to_json)
+      all = Record.all
+      expect(all).to be_kind_of Record
+      expect(all._proxy).to be_kind_of LHS::Collection
+      expect(all.count).to eq 0
```
pape-io: Why 0? Shouldn't it be 100, as there are 100 items in the body?

ryabrody: True. Sorry.

spec/record/all_spec.rb
ryabrody: True. Sorry.

ryabrody:
@pape-io It seems to work i tested it in the location app inside the branch called "num-list-sex-219". But there are still some red specs in this PR. Can u pls take care of them? Thx very mucho.

PR #90, opened 2016-03-02
Fix includes

Averethel:
`includes` was not duplicating a class so it was always hanging around.
The issue become visible when we changed the `create` behaviour.
It is fixed now


pape-io:
:+1:

PR #93, opened 2016-03-07
+ configurable endpoints

pape-io:
- Adds configuration ability to each LHS:Record (see added documentation)
- This one allows to use endpoints with different keys for pagination (like search)

### DIFF
--- a/lib/lhs/collection.rb
+++ b/lib/lhs/collection.rb
@@ -7,17 +7,18 @@ class LHS::Collection < LHS::Proxy
   include InternalCollection
 
   delegate :select, to: :_collection
+  delegate :_record, to: :_data
 
   def total
-    _data._raw[:total]
+    _data._raw[_record.total_key]
   end
 
   def limit
-    _data._raw[:limit]
+    _data._raw[_record.limit_key]
   end
 
   def offset
-    _data._raw[:offset]
+    _data._raw[_record.offset_key]
   end
 
   def href
@@ -26,8 +27,8 @@ class LHS::Collection < LHS::Proxy
 
   def _collection
     raw = _data._raw if _data._raw.is_a?(Array)
-    raw ||= _data._raw[:items]
-    Collection.new(raw, _data, _data._record_class)
+    raw ||= _data._raw[_record.items_key]
+    Collection.new(raw, _data, _record)
   end
 
   delegate :_raw, to: :_data
--- a/lib/lhs/concerns/record/all.rb
+++ b/lib/lhs/concerns/record/all.rb
@@ -12,12 +12,10 @@ class LHS::Record
       # for the following pages and concatenate all the results in order to return
       # all the objects for a given resource.
       def all(params = {})
-        all = []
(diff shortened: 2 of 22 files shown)


lib/lhs/collection.rb
```diff
@@ -7,17 +7,18 @@ class LHS::Collection < LHS::Proxy
   include InternalCollection
 
   delegate :select, to: :_collection
+  delegate :_record, to: :_data
 
   def total
-    _data._raw[:total]
+    _data._raw[_record.total_key]
```
kot-begemot: Does underscore here stands for private attribute?

lib/lhs/concerns/record/endpoints.rb
```diff
@@ -31,7 +31,7 @@ def endpoint(url, options = nil)
 
       def for_url(url)
         return unless url
-        _template, record = LHS::Record::Endpoints.all.detect do |template, _record_class|
+        _template, record = LHS::Record::Endpoints.all.detect do |template, _record|
```
kot-begemot: Does the record is used here anywhere? if not then you can replace it with `_`

kot-begemot:
Two comments, and if rubocop is happy, i am is happy :+1:

lib/lhs/collection.rb
pape-io: oui

lib/lhs/concerns/record/endpoints.rb
pape-io: Good point. Will do.

lib/lhs/collection.rb
pape-io: Actually it means 'internal'. So it's not private. It's accessible, but shouldn't be used from consumers of the library.

lib/lhs/concerns/record/all.rb
```diff
@@ -12,12 +12,10 @@ module ClassMethods
       # for the following pages and concatenate all the results in order to return
       # all the objects for a given resource.
       def all(params = {})
-        all = []
-        default_max_limit = 100
-        data = request(params: params.merge(limit: default_max_limit))
-        all.concat(all_items_from(data))
-        request_all_the_rest(data, all, params) if data._raw.is_a?(Hash) && data._raw[:total]
-        data._record_class.new(LHS::Data.new(all, nil, self))
+        limit = params[limit_key] || 100
```
acoiro: magic number!

acoiro:
:+1:

lib/lhs/collection.rb
eduardschaeli: Not to open a can of worms here, but isn't that what ```protected``` is for?

pape-io: The idea of using some internals is to flag variables that 'could' be used from outside, but can change behaviour at anytime. As the project is getting more and more mature, we can consider moving to protected attributes :+1:.

pape-io: Those `internal` attributes also help exploring and understanding structures.

PR #98, opened 2016-03-31
Fix all where result

ryabrody:
# TODO
- [ ] merge this PR https://github.com/local-ch/lhs/pull/96 before reviewing this one
- [ ] merge this PR https://github.com/local-ch/lhs/pull/97 before reviewing this one

# What I did 
Remove adding +1 to set `offset` in `request_all_the_rest` the 100 record (index 101) got lost because of that (may not the right approach)

### DIFF
--- a/lib/lhs/concerns/record/all.rb
+++ b/lib/lhs/concerns/record/all.rb
@@ -16,7 +16,6 @@ class LHS::Record
       def all(params = {})
         limit = params[limit_key] || DEFAULT_LIMIT
         data = request(params: params.merge(limit_key => limit))
-        binding.pry
         request_all_the_rest(data, params) if data._raw.is_a?(Hash) && data._raw[total_key]
         data._record.new(LHS::Data.new(data, nil, self))
       end
@@ -35,8 +34,7 @@ class LHS::Record
         total_left = data._raw[total_key] - data.count
         limit = data._raw[limit_key] || data.count
         if limit > 0
-          requests = total_left / limit
-          binding.pry
+          requests = (total_left.to_f / limit).ceil
           requests.times do |i|
             offset = limit * (i + 1) + 1
             data._raw[items_key].concat all_items_from request(
--- a/lib/lhs/concerns/record/all.rb
+++ b/lib/lhs/concerns/record/all.rb
@@ -16,6 +16,7 @@ class LHS::Record
       def all(params = {})
         limit = params[limit_key] || DEFAULT_LIMIT
         data = request(params: params.merge(limit_key => limit))
+        binding.pry
         request_all_the_rest(data, params) if data._raw.is_a?(Hash) && data._raw[total_key]
         data._record.new(LHS::Data.new(data, nil, self))
       end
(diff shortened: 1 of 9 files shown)


spec/collection/configurable_spec.rb
```diff
@@ -29,7 +29,7 @@ class Search < LHS::Record
       endpoint ':search/:type'
     end
     stub_request(:get, "http://local.ch/search/phonebook?size=10").to_return(body: first_response_data.to_json)
-    stub_request(:get, "http://local.ch/search/phonebook?size=10&start=11").to_return(body: second_response_data.to_json)
+    stub_request(:get, "http://local.ch/search/phonebook?size=10&start=10").to_return(body: second_response_data.to_json)
```
pape-io: There are to different ways for BE-endpoints to implement pagination. Either with allowing the client to paginating providing the current page: 1,2,3,4,5 or by allowing the client to provide the current offset: 1,6,11 etc.

This example and the current pagination feature (all) was implemented for a BE-endpoint that was using offset.
If you are now faced with a BE-endpoint that just works with pages, we have to consider making the client (LHS) smarter to be able to either use pages (1,2,3) or offset (1,6,11)

spec/record/all_spec.rb
```diff
@@ -15,10 +15,10 @@ class Record < LHS::Record
     it 'fetches all records from the backend' do
       stub_request(:get, "#{datastore}/feedbacks?limit=100")
         .to_return(status: 200, body: { items: (1..100).to_a, total: 300, limit: 100, offset: 0 }.to_json)
-      stub_request(:get, "#{datastore}/feedbacks?limit=100&offset=101")
-        .to_return(status: 200, body: { items: (101..200).to_a, total: 300, limit: 100, offset: 101 }.to_json)
-      stub_request(:get, "#{datastore}/feedbacks?limit=100&offset=201")
-        .to_return(status: 200, body: { items: (201..300).to_a, total: 300, limit: 100, offset: 201 }.to_json)
+      stub_request(:get, "#{datastore}/feedbacks?limit=100&offset=100")
```
pape-io: Page vs. Offset: https://github.com/local-ch/lhs/pull/98/files#diff-d9bdab362bb9ea000ba39992aa3d980fR32

pape-io: With limit off 100 and the first item is 1 you would receive item 1-100 from the BE already with the first request.
So the second page will start with item 101, but we will talk about that.

lib/lhs/concerns/record/all.rb
```diff
@@ -34,9 +34,9 @@ def request_all_the_rest(data, params)
         total_left = data._raw[total_key] - data.count
         limit = data._raw[limit_key] || data.count
         if limit > 0
-          requests = total_left / limit
+          requests = (total_left.to_f / limit).ceil
```
pape-io: Thats actually the key improvement of this PR. Together with https://github.com/local-ch/lhs/pull/98/files#diff-a26e7bc8dd017662e8b5a58f832ffce8R68

We should consider making a function out of `total_left` and also move the logic there. Furthermore I think we should ensure that `total_left`, provided by this function, should always return a natural number (1,2,3,4,5 etc.).

pape-io:
Superseded by https://github.com/local-ch/lhs/pull/102

PR #101, opened 2016-04-06
! navigating nested data for self initalized records

pape-io:


lib/lhs/collection.rb
```diff
@@ -47,6 +47,11 @@ def respond_to_missing?(name, include_all = false)
 
   private
 
+  def items_key
+    return _record.items_key if _record
```
acoiro: inline? `_record ? _record.items_key : :times`

acoiro:
otherwise ok :+1:

PR #102, opened 2016-04-07
Improve working with paginated endpoints/apis

pape-io:
LHS does pagination when you request `all`elements from the backend.

Before:

LHS was only supporting one pagination strategy, called `start`.
*example*
```ruby
request1(limit: 100, start: 1) -> [1..100]
request2(limit: 100, start: 101) -> [101..200]
etc.
```

New: 

LHS supports now 3 pagination strategies: `start`, `offset` and `page`.

*example start-strategy* -> see previous example

*example offset-strategy*
```ruby
request1(limit: 100, offset: 0) -> [1..100]
request2(limit: 100, offset: 100) -> [101..200]
etc.
```

*example page-strategy*
```ruby
request1(limit: 100, page: 1) -> [1..100]
request2(limit: 100, page: 2) -> [101..200]
etc.
```


lib/lhs/pagination.rb
```diff
@@ -0,0 +1,101 @@
...
+    offset
+  end
+
+  def next_offset
+    current_page + 1
+  end
+
+end
+
+class StartPagination < Pagination
```
skaestle: not a big fan of having multiple classes in the same file. not too worried in this case tho as these classes clearly should only be used together. maybe it would be worthwile to spend some time to see if that signal could be sent differernlty to the user of the class. Maybe modules or concerns? not sure

spec/record/all_spec.rb
```diff
@@ -1,68 +0,0 @@
-require 'rails_helper'
```
skaestle: i suppose these tests moved to paginatable_collection_spec?

spec/record/pagination_spec.rb
```diff
@@ -56,5 +56,14 @@ class Feedback < LHS::Record
     it 'responds to next_page' do
       expect(record.next_page).to eq(next_page)
     end
+
+    context 'when total of pages is not diviable trough limit' do
```
skaestle: typo `dividable` & `through` (should actually be `by`)

skaestle:
me likey

lib/lhs/pagination.rb
pape-io: It's actually [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern) by example and pretty much used that way.

Modules and concerns wouldn't allow separated classes and object-oriented separation. It would just lead to the if-else-nightmare-antipattern.

spec/record/all_spec.rb
pape-io: oui

ryabrody:
@pape-io Maybe this is out of scope but here is a suggestion for a imporved description in the readme for the "Partial Kaminari support"-section:


## Partial Kaminari support

LHS implements an interface that makes it partially working with Kaminari. 
The kaminari’s page parameter is in `params [:page]`. For example, you can use kaminari to render paginations based on LHS Records. Typically, your controller code will look like this:

```ruby
# controller
params[:page] = 0 if params[:page].nil?
page = params[:page].to_i
limit = 100
offset = (page - 1) * limit
@items = Record.where({ limit: limit, offset: offset }))
```

```ruby
# view
= paginate @items
```

pape-io:
Took your suggestion @ryabrody

spec/record/pagination_spec.rb
pape-io: fixed

Issue #103, opened 2016-04-07
LHS::WhereChain


pape-io:
```
Callcenters.where.callcenters_limited_to(10).where(type: 'Validated')
def callcenters_limited_to(limit)
    where(limit: limit)
end
```

Decide if dataset has to be loaded:
```
!(??????.new.methods + LHS::Record + LHS::RecordChain.new.methods).include?(:where)
```

pape-io:
We start with `where` , maybe we add stuff like `find_by` etc. later.
We don't think that `all` has to be chainable.

pape-io:
Implemented here: https://github.com/local-ch/lhs/pull/107

PR #107, opened 2016-05-04
+ where chains

pape-io:
**As lazy evaluated where statements can make tests fail in consuming application I will do a major version increase after merge**

Solves: https://github.com/local-ch/lhs/issues/103

## Scopes / WhereChains

LHS supports WhereChains. That allows you to chain multiple where-queries and helps you organizing query parameters that are used often in scopes (methods):

```ruby
class Record < LHS::Record
  endpoint 'records/'
  endpoint 'records/:id'

  def self.blue
    where(color: 'blue')
  end
end

records = Record.blue.where(available: true)
...
records.where(visible: true).each do |record|
  ...
end
```
The example would fetch the following in the end: `{color: blue, available: true, visible: true}`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,28 @@ This uses the `:datastore/v2/feedbacks` endpoint, cause `:campaign_id` was not p
 
 Uses the `:datastore/v2/content-ads/:campaign_id/feedbacks` endpoint.
 
+## Scopes / WhereChains
+
+LHS supports WhereChains. That allows you to chain multiple where-queries and helps you organizing query parameters that are used often in scopes (methods):
+
+```ruby
+class Record < LHS::Record
+  endpoint 'records/'
+  endpoint 'records/:id'
+
+  def self.blue
+    where(color: 'blue')
+  end
+end
+
+records = Record.blue.where(available: true)
+...
+records.where(sort: 'DESC').each do |record|
+  ...
+end
+```
+The example would fetch the following in the end: `{color: blue, available: true, sort: 'DESC'}`.
+
 ## Find single records
 
 `find` finds a unique record by uniqe identifier (usualy id).
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ The example would fetch records with the following parameters: `{color: blue, av
 
 ## Where values hash
 
-Returns a hash of where conditions (active-record-like).
+Returns a hash of where conditions.
 Common to use in tests, as where queries are not performing any HTTP-requests when no data is accessed.
 
 ```ruby
--- a/README.md
+++ b/README.md
@@ -93,8 +93,8 @@ In order to make common where statements reusable you can organise them in scope
 class Record < LHS::Record
   endpoint 'records/'
   endpoint 'records/:id'
-  scope :blue, -> {color: 'blue'}
-  scope :available ->(state) {available: state}
+  scope :blue, -> { where(color: 'blue') }
+  scope :available ->(state) { where(available: state) }
 end
 
 records = Record.blue.available(true)
--- a/README.md
+++ b/README.md
@@ -95,7 +95,9 @@ records = Record.where(color: 'blue').where(available: true).where(color: 'red')
 
 expect(
   records
-).to have_requested "something" # will fail as no http request is made (no data requested)
+).to have_requested(:get, %r{records/})
+  .with(query: hash_including(color: 'blue', available: true))
+# will fail as no http request is made (no data requested)
 
 expect(
   records.where_values_hash
--- a/README.md
+++ b/README.md
@@ -66,27 +66,40 @@ This uses the `:datastore/v2/feedbacks` endpoint, cause `:campaign_id` was not p
 
 Uses the `:datastore/v2/content-ads/:campaign_id/feedbacks` endpoint.
 
-## Scopes / WhereChains
+## Chaining where statements
 
-LHS supports WhereChains. That allows you to chain multiple where-queries and helps you organizing query parameters that are used often in scopes (methods):
+LHS supports chaining where statements. 
+That allows you to chain multiple where-queries:
 
 ```ruby
 class Record < LHS::Record
   endpoint 'records/'
   endpoint 'records/:id'
-
-  def self.blue
-    where(color: 'blue')
-  end
 end
 
-records = Record.blue.where(available: true)
+records = Record.where(color: 'blue')
 ...
-records.where(visible: true).each do |record|
+records.where(available: true).each do |record|
   ...
 end
 ```
-The example would fetch the following in the end: `{color: blue, available: true, visible: true}`.
+The example would fetch records with the following parameters: `{color: blue, available: true}`.
+
+## Scopes: Reuse where statements
+
+In order to make common where statements reusable you can organise them in scopes:
+
+```ruby
+class Record < LHS::Record
+  endpoint 'records/'
+  endpoint 'records/:id'
+  scope :blue, -> {color: 'blue'}
+  scope :available ->(state) {available: state}
+end
+
+records = Record.blue.available(true)
+The example would fetch records with the following parameters: `{color: blue, visible: true}`.
+```
 
 ## Find single records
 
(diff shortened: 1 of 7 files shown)


lib/lhs/concerns/record/where.rb
```diff
@@ -5,11 +5,51 @@ class LHS::Record
   module Where
     extend ActiveSupport::Concern
 
+    class WhereChain
```
gonzalocasas: instead of `WhereChain`, maybe `Predicate` is a nicer name?

pape-io: https://github.com/rails/rails/blob/ad078767fb4d4a78224efaaf541e9bc847cde245/activerecord/lib/active_record/relation/query_methods.rb#L13

Where do you get `Predicate` from? We usually try to stick to `ActiveRecord` concepts and naming. That's why we would like to call it WhereChains. It's also tightly coupled to where-statemenets.

README.md
```diff
@@ -66,6 +66,28 @@ This uses the `:datastore/v2/feedbacks` endpoint, cause `:campaign_id` was not p
...
+  endpoint 'records/:id'
+
+  def self.blue
+    where(color: 'blue')
+  end
+end
+
+records = Record.blue.where(available: true)
+...
+records.where(sort: 'DESC').each do |record|
```
gonzalocasas: why is the example of `where` using `sort`? Semantically, you should support `OrderBy` chaining as a separate operation, because `orderby` needs to keep the chain order, cannot do a `deep_merge` as this `where` predicate construction does.

lib/lhs/concerns/record/where.rb
gonzalocasas: ok, didn't know. I'm not that familiar with `ActiveRecord` terminology. So, in this case, stick to that. `Predicate` is used in other languages to define the same construct.

README.md
acoiro: I second that, I would expect a "filter" when using `.where`

acoiro:
I guess we don't need to handle the case of multiple `where` for the same key.

lib/lhs/concerns/record/where.rb
```diff
@@ -5,11 +5,51 @@ class LHS::Record
...
+        resolve.send(name, *args, &block)
+      end
+
+      def respond_to_missing?(name, include_all = false)
+        resolve.respond_to?(name, include_all)
+      end
+
+      def resolve
+        return @resolved if @resolved
+        @resolved = @record.new(
```
acoiro: `@resolved ||= ...` should give the same

README.md
pape-io: I will change the example and replace the 'sort' part. OrderBy is not treated in any specific manner, yet. (Maybe we will add that later).

lib/lhs/concerns/record/where.rb
pape-io: true

pape-io:
@acoiro multiple wheres for the same key are handled: Last one wins.

README.md
pape-io: @gonzalocasas I dont see the 'chain order' difference in the order by case:

```
records = Record.where(color: blue, sort: 'ASC')
records = records.where(sort: 'DESC')
```
Last one wins!

gonzalocasas: Multiple sort criteria. For example, if you want to specify sort by `name` and then by `creation_date`, the hash is not order-preserving, so you cannot know which sorting condition goes first. In any case, implementing multiple sort criteria probably means knowing more about the endpoint that you can using REST alone.

pape-io: @gonzalocasas now I got it. I will open an issue with that case ;)

spec/record/where_chains_spec.rb
```diff
@@ -29,6 +29,16 @@ def uppercase_name
...
     end
+
+    it 'takes the last value for chains with same name parameters' do
+      stub_request(:get, "http://datastore/v2/records/?parameter=last")
+          .to_return(body: [{ name: 'Steve' }])
+      records = Record.where(parameter: 'first').where(parameter: 'last')
+      # initiate the call, the stub will fail if the wrong parameter 'won'
+      records.first
+    end
+    # TODO: add a test that resolves to method_missing
```
skaestle: @pape-io as requested

README.md
```diff
@@ -66,6 +66,41 @@ This uses the `:datastore/v2/feedbacks` endpoint, cause `:campaign_id` was not p
...
+
+## Scopes: Reuse where statements
+
+In order to make common where statements reusable you can organise them in scopes:
+
+```ruby
+class Record < LHS::Record
+  endpoint 'records/'
+  endpoint 'records/:id'
+  scope :blue, -> {color: 'blue'}
```
ryabrody: put a space before/after curly brackets

README.md
```diff
@@ -66,6 +66,41 @@ This uses the `:datastore/v2/feedbacks` endpoint, cause `:campaign_id` was not p
...
+  scope :available ->(state) {available: state}
```
ryabrody: put a space before/after curly brackets

README.md
```diff
@@ -66,6 +66,41 @@ This uses the `:datastore/v2/feedbacks` endpoint, cause `:campaign_id` was not p
...
+end
+
+records = Record.blue.available(true)
+The example would fetch records with the following parameters: `{color: blue, visible: true}`.
```
ryabrody: put a space before/after curly brackets

README.md
```diff
@@ -66,6 +66,58 @@ This uses the `:datastore/v2/feedbacks` endpoint, cause `:campaign_id` was not p
...
+end
+```
+The example would fetch records with the following parameters: `{color: blue, available: true}`.
+
+## Where values hash
+
+Returns a hash of where conditions (active-record-like).
+Common to use in tests, as where queries are not performing any HTTP-requests when no data is accessed.
+
+```
```
ryabrody:
```ruby

File: README.md [hunk 9]
```diff
@@ -85,6 +85,23 @@ end
 ```
 The example would fetch records with the following parameters: `{color: blue, available: true}`.
 
+## Where values hash
+
+Returns a hash of where conditions (active-record-like).
```
ryabrody: I would not add this: `(active-record-like)`. I think you never used this in the doc before?

README.md
```diff
@@ -66,6 +66,58 @@ This uses the `:datastore/v2/feedbacks` endpoint, cause `:campaign_id` was not p
...
+Common to use in tests, as where queries are not performing any HTTP-requests when no data is accessed.
+
+```ruby
+records = Record.where(color: 'blue').where(available: true).where(color: 'red')
+
+expect(
+  records
+).to have_requested "something" # will fail as no http request is made (no data requested)
```
ryabrody: I would rather write
```ruby
expect(
  records
).to have_requested(:get, %r{records/})
  .with(query: hash_including(color: 'blue', available: true))
# will fail as no http request is made (no data requested)
```
than `"something"` .

README.md
```diff
@@ -66,6 +66,60 @@ This uses the `:datastore/v2/feedbacks` endpoint, cause `:campaign_id` was not p
...
+...
+records.where(available: true).each do |record|
+  ...
+end
+```
+The example would fetch records with the following parameters: `{color: blue, available: true}`.
+
+## Where values hash
+
+Returns a hash of where conditions (active-record-like).
```
ryabrody: Do you really wanna use "(active-record-like)" ?

spec/record/where_chains_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    it 'resolves triggered by method missing' do
+      expect(records._raw).to eq [{ name: 'Steve' }]
+      expect(
+        Record.where(color: 'blue').where(range: '>26', available: true).first.name
+      ).to eq 'Steve'
+    end
+  end
+
+  context 'multiple paramters' do
```
skaestle: small typo `multiple paramter` vs `multiple parameter`

ryabrody:
just a small cmnt otherwise 👍

spec/record/where_chains_spec.rb
pape-io: fixed

PR #112, opened 2016-05-12
Keys method spec file

kot-begemot:
Key method for record returns nil, while it should be array of top-level keys.


Averethel:
@kot-begemot @pape-io what's the status here?

PR #113, opened 2016-05-17
! allow to load classes twice: ignore duplicate endpoints, they are n…

pape-io:
Some consuming applications load the Record classes multiple times. In those cases LHS always raises the `clashing endpoints` error, as it does not ignore real duplicates. We can ignore real duplicates, as the clashing should prevent different endpoints of the same Record to contain the same placeholders!

### DIFF
--- a/lib/lhs/concerns/record/endpoints.rb
+++ b/lib/lhs/concerns/record/endpoints.rb
@@ -47,9 +47,13 @@ class LHS::Record
       end
 
       # Prevent clashing endpoints.
-      def sanity_check(endpoint)
-        placeholders = endpoint.placeholders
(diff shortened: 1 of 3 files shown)


skaestle:
not quite sure I get which problem this PR solves

pape-io:
@skaestle updated the above text. I hope it gets clearer now?
>
Some consuming applications load the Record classes multiple times. In those cases LHS always raises the `clashing endpoints` error, as it does not ignore real duplicates. We can ignore real duplicates, as the clashing should prevent different endpoints of the same Record to contain the same placeholders!

skaestle:
thanks!

kot-begemot:
Yes, that works now :+1:

PR #115, opened 2016-05-18
! nested data casting in collections

pape-io:
Fixes nested data casting in case of collections.

### DIFF
--- a/lib/lhs/concerns/collection/internal_collection.rb
+++ b/lib/lhs/concerns/collection/internal_collection.rb
@@ -33,8 +33,9 @@ class LHS::Collection < LHS::Proxy
       private
 
       def cast_item(item)
+        record = LHS::Record.for_url(item[:href]) if item[:href]
         data = LHS::Data.new(item, @parent, @record)
-        return @record.new(data) if @record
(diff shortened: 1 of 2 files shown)


pape-io:
@cedricwider

lib/lhs/concerns/collection/internal_collection.rb
```diff
@@ -33,8 +33,9 @@ def each(&_block)
       private
 
       def cast_item(item)
+        record = LHS::Record.for_url(item[:href]) if item[:href]
         data = LHS::Data.new(item, @parent, @record)
-        return @record.new(data) if @record
+        return (@record || record).new(data) if @record || record
```
acoiro: quite hard to understand the difference between `@record` and `record`. Better naming would improve readability.

acoiro:
otherwise 👍

lib/lhs/concerns/collection/internal_collection.rb
pape-io: Giving better names actually made me realise that it should be the other way around... 🙈

cedricwider:
👍

PR #117, opened 2016-05-19
+ add options to the chain

pape-io:
## Request based options

You can apply options to the request chain. Those options will be forwarded to the request perfomed by the chain/query.

```ruby
  options = { auth: { bearer: '123456' } }
  
  authenticated_record = Record.options(options)
  
  blue_records = authenticated_record.where(color: 'blue')
  active_records = authenticated_record.where(active: true)

  authenticated_record.create(color: 'red')
  
  record = authenticated_record.find(123)
  record.name = 'Walter'
  record.valid?(options)
  record.save(options)
  record.destroy(options)
  record.update({ name: 'Steve' }, options)
```


acoiro:
If this is an attempt to fix the missing access token issues, then this isn't enough: for some reason the failures happen only on POST, PATCH & DELETE. @skaestle was saying that it could be a server config which retries these requests in same cases, but I am not sure.

acoiro:
otherwise 👍

skaestle:
here is some context: https://trac.nginx.org/nginx/ticket/488

appareantly they have a fixed version available.

pape-io:
Will also apply it to `create`, `save`, `destroy`, `update` and `validate`

README.md
```diff
@@ -156,6 +156,28 @@ If no record is found, `nil` is returned.
...
+  
+  blue_records = authenticated_record.where(color: 'blue')
+  active_records = authenticated_record.where(active: true)
+
+  authenticated_record.create(color: 'red')
+  
+  record = authenticated_record.find(123)
+  record.name = 'Walter'
+  record.valid?(options)
+  record.save(options)
```
acoiro: it feels wrong to pass again the options to the `authenticated_record`. I hope we can fix the threading issue without merging it.

acoiro:
it does the job 👍 I would wait a bit before merging it, maybe we can fix it in another way without hacking LHS

pape-io:
@acoiro it's the other way around. We hacked CC with RequestStore, because this here, was not available back then. 

As I said before, I wanted to implement request based options for LHS anyhow.

README.md
pape-io: It's the opposite. It would be wrong not passing it again. Think about oAuth where every request authenticates itself!

The request happen isolated, so it's not ensured that they share the same token!

pape-io: If you want to configure options globally, you do so in the Record not on the Request level!

```
class Record < LHS::Record
  endpoint 'contracts/', auth: { bearer: '123 }
end
```

pape-io: And it's actually the same behaviour as the ActiveRecord chains. As they get resolved with `find` in the example, no chain meta data is allowed to influence upcoming queries after resolving!

pape-io: `valid?` and `save` as in the example, start new chains.

acoiro:
Hmm really? We loose a bit of abstraction with these `options` param

pape-io:
@acoiro can you elaborate on loosing abstraction? it's optional.

acoiro:
Without `options` for auth it was like using a local database, no details about the internals. Not a big deal anyway.

pape-io:
I mean, it will still just be used for edge cases in the future, like oauth. But it's good to have it, as we would have implemented the bearer token forwarding with it. I still don't like the Thread approach, even if it seems to work now, it's against separation of concerns!

acoiro:
I see, however it isn't really an edge case: every call will need the `options` since every call will need oauth.

pape-io:
@acoiro We could add something that hooks into the controller before and authenticates configured Records for you.
That way you wouldn't have to deal with options while developing, you have authenticated records, and you dont bypass MVC. 

What do you think?

pape-io:
But this idea would be based on the current stuff anyway, as it would be just a little bit more sugar.

pape-io:
```ruby
class Record < LHS::Record
  endpoint 'contracts'
  endpoint 'contracts/:id'

  before_controller ->{
    Record.options(auth: { bearer: access_token })
  }
end
```

And nothing about options anywhere else.
What do you think @acoiro ?

pape-io:
As discussed, I will do the `before_controller` thingy in a separate PR.

Things I change before I merge this one here:
```
  Record.options(options).save
  Record.options(options).valid?
  Record.options(options).destroy(data)
  Record.options(options).update(data)
```
Instead of
```
  Record.save(options)
  Record.valid?(options)
  Record..destroy(options)
  Record.update(data, options)
```

acoiro:
👍

PR #119, opened 2016-05-23
Prevent error while loading missing includes

cedricwider:
Consider the followig response:
```javascript
{
 ...
 "items" : [
    {
      "some_field" : "some value",
      "some_object" : {
        "href" : "...some url..."
    }, {
      "some_field" : "some other value"
    }
  ]
}
```

The following statement raises an error, because in the second items-object 'some_object' is not defined.
```ruby
TheRecord.includes(:some_object).all
```

This change addresses the fact that sometimes servers omit sending back objects in the response when they're not present. (null)


lib/lhs/version.rb
```diff
@@ -1,3 +1,3 @@
 module LHS
-  VERSION = "5.1.0"
+  VERSION = "5.1.1"
```
pape-io: Versioning happens on master

lib/lhs/concerns/record/request.rb
```diff
@@ -92,14 +103,31 @@ def merge_explicit_params!(params)
...
-        responses = LHC.request(options)
-        data = responses.map { |response| LHS::Data.new(response.body, nil, self, response.request) }
-        data = LHS::Data.new(data, nil, self)
-        handle_includes(including, data) if including
+        requests = options.map do |option|
+          next unless option.present?
+          process_options(option, find_endpoint(option[:params]))
+        end
+        data = LHC.request(requests.compact).map { |response| LHS::Data.new(response.body, nil, self, response.request) }
+        data = restore_with_nils(data, locate_nils(requests)) # nil objects in data provide location information for mapping
```
pape-io: keep this line included since data is overwritten later on

lib/lhs/concerns/record/request.rb
```diff
@@ -38,10 +39,14 @@ def convert_option_to_endpoints(option)
 
       # Extends existing raw data with additionaly fetched data
       def extend_raw_data(data, addition, key)
-        if data._proxy.is_a? LHS::Collection
+        return if addition.empty?
+        if data.collection?
           data.each_with_index do |item, i|
             item = item[i] if item.is_a? LHS::Collection
-            item._raw[key.to_sym].merge!(addition[i]._raw)
+            placeholder = item._raw[key.to_sym]
```
pape-io: `original` or `link`

lib/lhs/concerns/record/request.rb
```diff
@@ -38,10 +39,14 @@ def convert_option_to_endpoints(option)
...
+            if placeholder.present?
+              placeholder.merge!(addition[i]._raw)
```
pape-io: would post condition one liner here

lib/lhs/concerns/record/request.rb
```diff
@@ -38,10 +39,12 @@ def convert_option_to_endpoints(option)
...
       # Extends existing raw data with additionaly fetched data
       def extend_raw_data(data, addition, key)
-        if data._proxy.is_a? LHS::Collection
+        return if addition.empty?
+        if data.collection?
           data.each_with_index do |item, i|
             item = item[i] if item.is_a? LHS::Collection
-            item._raw[key.to_sym].merge!(addition[i]._raw)
+            link = item._raw[key.to_sym]
+            link.merge!(addition[i]._raw) if link.present?
```
pape-io: Could you try to use
```
  item.merge_raw!(addition[i])
```
From LHS::Data https://github.com/local-ch/lhs/blob/prevent_error_while_loading_missing_includes/lib/lhs/data.rb
Would simplify this part here, would use the method made for link data merging.

You might have to move some `presence/present?` logic over there.

lib/lhs/concerns/record/request.rb
```diff
@@ -59,19 +62,25 @@ def handle_includes(includes, data)
       end
 
       def handle_include(included, data, sub_includes = nil)
-        return unless data.present?
+        return unless data.present? && load_includes?(data, included)
```
pape-io: could you switch from `unless` to `if` here. Me personally I find `unless` in combination with `&&` confusing.

lib/lhs/concerns/record/request.rb
```diff
@@ -59,19 +62,25 @@ def handle_includes(includes, data)
...
         options =
-          if data._proxy.is_a? LHS::Collection
-            options_for_multiple(data, included)
+          if data.collection?
```
pape-io: +1

lib/lhs/concerns/record/request.rb
```diff
@@ -92,14 +101,31 @@ def merge_explicit_params!(params)
       end
 
       def multiple_requests(options)
-        options = options.map { |option| process_options(option, find_endpoint(option[:params])) }
-        responses = LHC.request(options)
-        data = responses.map { |response| LHS::Data.new(response.body, nil, self, response.request) }
-        data = LHS::Data.new(data, nil, self)
-        handle_includes(including, data) if including
+        requests = options.map do |option|
```
pape-io: I would stick with `options` as before, because you pass a set of options, which result in a set of requests, wich result in a set of responses. What do you think?

pape-io: I would stick with options as before, because you pass a set of options, which result in a set of requests, wich result in a set of responses. What do you think?

lib/lhs/concerns/record/request.rb
```diff
@@ -92,14 +101,31 @@ def merge_explicit_params!(params)
...
       end
 
+      def locate_nils(array)
+        nils = []
+        array.each_with_index { |val, i| nils << i if val.nil? }
+        nils
+      end
+
+      def restore_with_nils(array, nils)
+        arr = array.dup
```
pape-io: Principle of visibility and symmetrie violated, please call it `array`. Again shadowing in this particular case is actually a benefit!

lib/lhs/concerns/record/request.rb
```diff
@@ -92,14 +101,31 @@ def merge_explicit_params!(params)
...
+        unless data.empty?
+          data = LHS::Data.new(data, nil, self)
+          handle_includes(including, data) if including
+        end
         data
       end
 
+      def locate_nils(array)
+        nils = []
+        array.each_with_index { |val, i| nils << i if val.nil? }
```
pape-io: Symmetrie and visibility violated, please call it `index` and `value`

cedricwider: use value instead of val and index instead of i

lib/lhs/concerns/record/request.rb
cedricwider: shadowing array is okay (preferred) here.

lib/lhs/concerns/record/request.rb
```diff
@@ -38,10 +39,12 @@ def convert_option_to_endpoints(option)
 
       # Extends existing raw data with additionaly fetched data
       def extend_raw_data(data, addition, key)
-        if data._proxy.is_a? LHS::Collection
+        return if addition.empty?
+        if data.collection?
```
pape-io: BTW: collection? is not implemented yet ;) so this will also be `false`.

Data responds to everything with nil!

pape-io: I saw that you implemented it in here, I will take it for a separate PR. thx

pape-io:
Looks good to go 👍 let's merge, version and release

PR #121, opened 2016-05-24
+ methods to navigate data

pape-io:
+ `collection?` to check if data is a collection
+ `item?` to check if data is an item
+ `parent` to navigate up
+ equality for data and record in order to make things comparable

! not bypassing data anymore when forwarding data from record to proxy
! keeping the special case for rubys native select (which has to be solved with delegation)


cedricwider:
👍

Averethel:
👍

PR #122, opened 2016-05-25
Fix pagination string case

pape-io:
Fixes cases where pagination arguments have been passed as strings
```
Record.page(params[:page])
```
Params are usually strings.


ryabrody:
a 👍 for your changes. May you could also pls check my changes on the REAMDE?

pape-io:
@ryabrody 👍

PR #123, opened 2016-05-26
! immutable chains

pape-io:
Always returns a new chain, so that the original is not getting mutated:

```ruby
blue_records = Record.where(color: 'blue')

blue_and_active_records = blue_records.where(active: true)

```

### DIFF
--- a/lib/lhs/concerns/record/chainable.rb
+++ b/lib/lhs/concerns/record/chainable.rb
@@ -63,13 +63,15 @@ class LHS::Record
     class Chain
 
       # Instance exec is required for scope chains
-      delegated_methods = Object.instance_methods - [:instance_exec]
+      delegated_methods = Object.instance_methods - [:instance_exec, :clone]
       delegate(*delegated_methods, to: :resolve)
+
+      attr_accessor :_links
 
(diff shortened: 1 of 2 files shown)


lib/lhs/concerns/record/chainable.rb
```diff
@@ -180,20 +182,21 @@ def resolve
       private
 
       def push(link)
-        @chain += [link].compact
-        self
+        clone = self.clone
```
eduardschaeli: Is a shallow clone enough here? http://ruby-doc.org/core-2.3.1/Object.html#method-i-clone

pape-io: Sure:
```ruby
{ # Chain
  _links: [
    { # Link
      hash: parameters
    }
  ]
}
```

Only `_links` are getting mutated. The hashes inside the links are not getting mutated!
You agree @eduardschaeli ?

Averethel:
:+1:

lib/lhs/concerns/record/chainable.rb
eduardschaeli: Ah I see.

PR #126, opened 2016-05-31
+ includes with nested items

pape-io:
Enables including nested items inside arrays:
```
# place
{
  category_relations: [
    { href: "http://datastore/category/2" }
  ]
}
```

```ruby
Place.includes(:category_relations)
# fetches now nested items as well
```


lib/lhs/concerns/record/request.rb
```diff
@@ -46,6 +46,10 @@ def extend_raw_data(data, addition, key)
             link = item[key.to_sym]
             link.merge_raw!(addition[i]) if link.present?
           end
+        elsif data[key]._raw.is_a? Array
+          data[key].each_with_index do |item, index|
+            item._raw.merge!(addition[index]._raw)
+          end
```
Averethel: how about:
```ruby
data[key].zip(additions).each do |item, addition|
  item._raw.merge(addition._raw)
```

Averethel:
small comment otherwise 👍

lib/lhs/concerns/record/request.rb
pape-io: I have to modify the originals??? so how does this help @Averethel ?

pape-io: missing a `!`

Averethel:
```ruby
a = {}
b = {b: 1}
[a].zip([b]).each{ |a, b| a.merge!(b) }

a # =>  {b: 1}
```

PR #127, opened 2016-05-31
! improve data inspect

pape-io:
Data on the terminal now looks directly like raw data:

```ruby
data = Record.find(1)
data # { name: 'Steve' }
```

Before: 
```ruby
data
#<LHS::Data:0x007fe67a64f600 @_raw={:name=>\"Steve\"}, @_parent=nil, @_record=Record, @_proxy=#<LHS::Item:0x007fe67a64f420 @_data={:name=>\"Steve\"}, @_loaded=false>, @_request=nil, @_endpoint=nil>
```

On one hand this was leaking internals to the outside, second it makes people use `_raw` all over the place, but they shouldn't.

For internal reasons `inspect_without_focus` stays around.

### DIFF
--- a/lib/lhs/data.rb
+++ b/lib/lhs/data.rb
@@ -74,6 +74,11 @@ class LHS::Data
       _proxy.respond_to?(name, include_all)
   end
 
+  def inspect_with_focus
+    _raw
+  end
+  alias_method_chain :inspect, :focus
+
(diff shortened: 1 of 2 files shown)


Averethel:
👍

cedricwider:
👍

PR #131, opened 2016-06-02
Expand

pape-io:
LHS will ask the specific endpoint to `expand` the included linked resources on their side. If it detects that this didn't happen, it will fetch them itself.

### DIFF
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -165,15 +165,21 @@ class LHS::Record
       end
 
       # Prepare expand parameter in order to get expanding done by the endpoint
-      def handle_expands(expands)
+      def handle_expands(expands, prefix = nil)
         list = []
         if expands.is_a? Hash
-          expands.each { |expands, sub_expands|  }
-          fail 'pending'
+          expands.each do |key, sub_expands|
+            list << handle_expands(sub_expands, [prefix, key].compact.join('.'))
+          end
         elsif expands.is_a? Array
-          expands.each { |expands| handle_expands(expands) }
+          expands.each do |element|
+            list << handle_expands(element, prefix)
+          end
         else
-          expand.to_s
+          list << [
+            prefix,
+            expands
+          ].compact.join('.')
         end
         list.join(',')
       end
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -67,7 +67,7 @@ class LHS::Record
       end
 
       def handle_include!(included, data, sub_includes = nil)
-        return if data.blank? || skip_loading_includes?(data, included)
+        return if data.blank? || skip_loading_includes?(data, included) || expanded?(included, data, sub_includes)
         options =
           if data.collection?
             options_for_multiple(data, included)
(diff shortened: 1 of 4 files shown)


lib/lhs/concerns/record/request.rb
```diff
@@ -55,18 +55,19 @@ def extend_raw_data(data, addition, key)
...
         else
-          handle_include(includes, data)
+          handle_include!(includes, data)
         end
       end
 
-      def handle_include(included, data, sub_includes = nil)
-        return if data.blank? || skip_loading_includes?(data, included)
+      def handle_include!(included, data, sub_includes = nil)
+        return if data.blank? || skip_loading_includes?(data, included) || expanded?(included, data, sub_includes)
```
cedricwider: some might consider a statement like `return if a || b || c` do be worth refactoring out into a method. What do you think?

lib/lhs/concerns/record/request.rb
```diff
@@ -146,14 +153,37 @@ def options_for_nested_items(data, key)
 
       # Merge explicit params and take configured endpoints options as base
       def process_options(options, endpoint)
+        options = options.deep_dup
         options[:params].deep_symbolize_keys! if options[:params]
         options = (endpoint.options || {}).merge(options)
         options[:url] = compute_url!(options[:params]) unless options.key?(:url)
+        options[:params] ||= {}
```
cedricwider: moving this statement up could save you the `if options[:params]` on line 157

cedricwider:
added minor comments... other than that 👍

lib/lhs/concerns/record/request.rb
Averethel: I agree with @cedricwider - factoring out condition to method would be nice

lib/lhs/concerns/record/request.rb
```diff
@@ -146,14 +153,37 @@ def options_for_nested_items(data, key)
...
+        options[:params][:expand] = handle_expands(including) if including.present?
         merge_explicit_params!(options[:params])
         options.delete(:params) if options[:params] && options[:params].empty?
         options
       end
 
+      # Prepare expand parameter in order to get expanding done by the endpoint
+      def handle_expands(expands, prefix = nil)
+        list = []
```
Averethel: how about using `tap` here:
```ruby
[].tap do |list|
  ....
end.join(',')
```

lib/lhs/concerns/record/request.rb
```diff
@@ -146,14 +153,37 @@ def options_for_nested_items(data, key)
...
+          expands.each do |element|
+            list << handle_expands(element, prefix)
+          end
+        else
+          list << [
+            prefix,
+            expands
+          ].compact.join('.')
+        end
+        list.join(',')
```
Averethel: Should there not be also `compact` / empty removal?

Averethel:
Minor comments, once CI is happy so am I.

lib/lhs/concerns/record/request.rb
pape-io: Not necessary, as the necessary lines that add stuff to the list are compacting, but will add it for comprehensibility. We want to have a cleaned list.

lib/lhs/concerns/record/request.rb
pape-io: #tap that 👍

lib/lhs/concerns/record/request.rb
pape-io: done 👍

PR #132, opened 2016-06-06
Keep the headers passed via options to save

Averethel:
In the current implementation `save!` method always overwrites the headers passed via the options.
This PR fixes that misbehaviour.


pape-io:
Go ahead 👍 
Patch version increase on master!

spec/item/save_spec.rb
```diff
@@ -50,5 +50,17 @@ class Record < LHS::Record
         .to_return(status: 500)
       expect(-> { item.save! }).to raise_error LHC::ServerError
     end
+
+    it 'keeps header psassed in the options' do
```
skaestle: typo

PR #135, opened 2016-06-15
! skip loading empty includes

pape-io:
LHS was raising an exception when somebody tried to include empty collections.


acoiro:
👍

PR #136, opened 2016-06-15
! extend base items with collections

pape-io:
Fixes the extension of nested `items` while including nested resources

### DIFF
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -38,20 +38,45 @@ class LHS::Record
       end
 
       # Extends existing raw data with additionaly fetched data
-      def extend_raw_data(data, addition, key)
+      def extend_raw_data!(data, addition, key)
         return if addition.empty?
         if data.collection?
-          data.each_with_index do |item, i|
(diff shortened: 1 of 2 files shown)


lib/lhs/concerns/record/request.rb
```diff
@@ -38,23 +38,48 @@ def convert_option_to_endpoints(option)
       end
 
       # Extends existing raw data with additionaly fetched data
-      def extend_raw_data(data, addition, key)
+      def extend_raw_data!(data, addition, key)
```
kot-begemot: Do you plan to reintroduce backwards compatible method here?

kot-begemot:
A comment. As long as minor version is bumped it is :+1:

lib/lhs/concerns/record/request.rb
pape-io: It's a private method!?! why should I put it minor? in case it would introduce breaking changes it would need to be a major increase btw.. http://semver.org/

I just renamed the (internal) method and added a bang to emphasize it is changing the data you provide. Before that was not clear.

kot-begemot:
Then go for it :+1:

PR #137, opened 2016-07-06
! batch processing with configured pagination

pape-io:
Bug Fix

LHS was not capable to do batch processing (internal pagination) when configured LHS::Record had custom pagination configuration (like items = 'docs', total = 'totalResults' etc.).

This fixes it and uses configured pagination to be applied when doing batch processing.


cedricwider:
👍

PR #142, opened 2016-07-28
Forward options to included/referenced resources

pape-io:
In case you want to forward options for the requests beeing made to fetch included/referenced resources:
```
Favorite.includes(:place).references(place: { auth: { bearer: '123' }})
```

Major version increase ahead, because of new ruby > 2.0.0 dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,16 @@ Favorite.includes(:place).where(user_id: current_user.id)
 # Will include places and applies endpoint options to authenticate the request.
 ```
 
+### Forward options used for request made to include referenced resources
+
+Provide options to the requests made to include referenced resources:
+
+```
+
+  Favorite.includes(:place).references(place: { auth: { bearer: '123' }})
+
+```
+
 ## Map data
 
 To influence how data is accessed/provied, you can use mappings to either map deep nested data or to manipulate data when its accessed. Simply create methods inside the LHS::Record. They can access underlying data:
--- a/README.md
+++ b/README.md
@@ -174,6 +174,7 @@ After fetching [single](#find-single-records) or [multiple](#find-multiple-recor
 You can apply options to the request chain. Those options will be forwarded to the request perfomed by the chain/query.
 
 ```ruby
+  # Authenticate with OAuth
   options = { auth: { bearer: '123456' } }
   
   AuthenticatedRecord = Record.options(options)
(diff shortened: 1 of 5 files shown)


ryabrody:
rubocop ☝🏻

README.md
```diff
@@ -365,29 +366,39 @@ After include:
...
+This allows you to include protected resources (e.g. Basic auth) as endpoint options for oauth authentication get applied.
 
 The [Auth Inteceptor](https://github.com/local-ch/lhc-core-interceptors#auth-interceptor) from [lhc-core-interceptors](https://github.com/local-ch/lhc-core-interceptors) is used to configure the following endpoints.
 
 ```ruby
 class Favorite < LHS::Record
 
-  endpoint ':datastore/:user_id/favorites', auth: { bearer: -> { bearer_token } }
-  endpoint ':datastore/:user_id/favorites/:id', auth: { bearer: -> { bearer_token } }
+  endpoint ':datastore/:user_id/favorites', auth: { basic: { username: 'steve', password: 'can' } }
```
ryabrody: This looks like you are suggestion to put a non-encrypted password as parameter.

pape-io: @ryabrody You do, as the interceptors are build to solve common tasks for HTTP communication, encrypting is one of them, so you can give them non-encrypted passwords:
https://github.com/local-ch/lhc-core-interceptors/blob/master/lib/lhc-core-interceptors/auth.rb#L18
and they will encrypt it ;)

README.md
```diff
@@ -365,29 +366,39 @@ After include:
...
 # Will include places and applies endpoint options to authenticate the request.
 ```
 
+### Forward options used for request made to include referenced resources
+
+Provide options to the requests made to include referenced resources:
+
+```
+
+  Favorite.includes(:place).references(place: { auth: { bearer: '123' }})
```
ryabrody: Should the auth-stuff not going into the endpoint definiton?

pape-io: In case of OAuth, the token is only available on the controller level and here's why:
Rails loads all your models, when you application gets booted. It initializes/reinitializes the controller when a new request comes in. The OAuth token has to be requested and reevaluated for every single request. As MVC do not allow to share data between controllers and models, except if you forward them with explicit method calls on Models, and this is what you see here.

In case of basic auth, you're right, you define it on the endpoint-definition inside the Record, as those credentials are not changing during the request/response cycle.

lib/lhs/concerns/record/request.rb
```diff
@@ -80,17 +79,18 @@ def extend_base_item_with_collection!(data, addition, key)
...
 
-      def handle_includes(includes, data)
+      def handle_includes(includes, data, references = {})
+        references ||= {}
         if includes.is_a? Hash
-          includes.each { |included, sub_includes| handle_include(included, data, sub_includes) }
+          includes.each { |included, sub_includes| handle_include(included, data, sub_includes, references[included]) }
         elsif includes.is_a? Array
-          includes.each { |included| handle_includes(included, data) }
+          includes.each { |included| handle_includes(included, data, references[:included]) }
```
pape-io: Testcase missing for this.

cedricwider:
👍 after 1 on 1 review

PR #146, opened 2016-07-29
Expand included collection items

pape-io:
In the following case, LHS was not expanding included resources:

```
Customer.includes(:places).find(1)
GET 'customers/1'
```
```
{
  places: { href: 'customers/1/places' }
}
```
`includes` fetches places
```
GET 'customers/1/places'
{
  items: [
    { href: 'datastore/places/1' },
    { href: 'datastore/places/2' },
    { href: 'datastore/places/3' }
  ]
}
```
but this just includes a list of links, so LHS was not further including the requested resources.

Now it does also fetch all those links in order to fulfil the requested: `Customer.includes(:places).find(1)`

Backwards compatible: Minor

### DIFF
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -108,7 +108,7 @@ class LHS::Record
       def options_for_data(data, included = nil)
         if data.collection?
           options_for_multiple(data, included)
-        elsif data[included].collection?
+        elsif included && data[included].collection?
           options_for_nested_items(data, included)
         else
           url_option_for(data, included)
@@ -129,6 +129,7 @@ class LHS::Record
       end
 
       def no_expanded_data?(addition)
+        return false if addition.blank?
         if addition.item?
           (addition._raw.keys - [:href]).empty?
         elsif addition.collection?
@@ -197,7 +198,7 @@ class LHS::Record
 
       def options_for_nested_items(data, key = nil)
         data[key].map do |item|
-          url_option_for(item, key)
+          url_option_for(item)
         end
       end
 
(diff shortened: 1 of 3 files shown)


cedricwider:
👍

lib/lhs/concerns/record/request.rb
```diff
@@ -76,7 +76,14 @@ def extend_base_item_with_collection!(data, addition, key)
         if target._raw.is_a? Array
           data[key] = addition.map(&:_raw)
         else # hash with items
-          target._raw[items_key] = addition.map(&:_raw)
+          target._raw[items_key] ||= []
+          if target._raw[items_key].empty?
```
Averethel: We have some nesting here. Do you think it makes sense to extract this to auxiliary method?

lib/lhs/concerns/record/request.rb
```diff
@@ -92,16 +99,42 @@ def handle_includes(includes, data)
...
-            url_option_for(data, included)
-          end
+        options = options_for_data(data, included)
         addition = load_include(options, data, sub_includes)
         extend_raw_data!(data, addition, included)
+        expand_addition!(data, included) if no_expanded_data?(addition)
+      end
+
+      def options_for_data(data, included = nil)
+        if data.collection?
```
Averethel: I'd rather go for guard clauses:
```ruby
return options_for_multiple(data, included) if data.collection?
return options_for_nested_items(data, included) if included && data[included].collection?
url_option_for(data, included)
```
What do you think?

Averethel:
👍

lib/lhs/concerns/record/request.rb
pape-io: done 👍

lib/lhs/concerns/record/request.rb
pape-io: 👍

PR #149, opened 2016-08-11
Do not store values set by custom setters in _raw by default

Averethel:
Consider following use case:

```ruby
class Feedback < LHS::Record
  attr_accessor :listing
end

listing = Local::Listing(...)
f1 = Feedback.new(listing: listing)

f1._raw
  => { listing: listing }

f2 = Feedback.new.tap do |f|
  f.listing = listing
end

f2._raw
  => {}
```

This PR fixes this issue.

The reason it occurred was that we had an assumption that everything that has a custom setter will end up in the raw data anyway, which is not always true (think instance variables).

It wouldn't be a big of an issue if not for the recursive data structures, which then cause stack overflow when trying to send them.

### DIFF
--- a/lib/lhs/record.rb
+++ b/lib/lhs/record.rb
@@ -25,7 +25,7 @@ class LHS::Record
     data = LHS::Data.new({}, nil, self.class) unless data
     data = LHS::Data.new(data, nil, self.class) unless data.is_a?(LHS::Data)
     define_singleton_method(:_data) { data }
-    consider_custom_setters!
+    apply_custom_setters!
   end
 
   def as_json(options = nil)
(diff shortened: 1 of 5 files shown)


lib/lhs/record.rb
```diff
@@ -25,7 +25,7 @@ def initialize(data = nil)
     data = LHS::Data.new({}, nil, self.class) unless data
     data = LHS::Data.new(data, nil, self.class) unless data.is_a?(LHS::Data)
     define_singleton_method(:_data) { data }
-    consider_custom_setters
+    consider_custom_setters!
```
pape-io: might wanna call it just `customer_setters!` or `apply_custom_setters!` or `perform_custom_setters!`

Averethel: 👍  I'd go for `apply_custom_setters!`

pape-io:
after checking https://github.com/local-ch/lhs/pull/150 👍

PR #151, opened 2016-08-25
Find in parallel + Handle Errors (chained)

pape-io:
# Find multiple single records in parallel

In case you want to fetch multiple records by id in parallel, you can also do this with `find`:

```ruby
Record.find(1, 2, 3)
```

If you want to inject values for the failing records, that might not have been found, you can inject values for them with error handlers:

```ruby
data = Record
  .handle(LHC::Unauthorized, ->(response) { Record.new(name: 'unknown') })
  .find(1, 2, 3)
data[1].name # 'unknown'
```

### DIFF
--- a/lib/lhs/concerns/record/chainable.rb
+++ b/lib/lhs/concerns/record/chainable.rb
@@ -162,7 +162,10 @@ class LHS::Record
       end
 
       def find(args)
-        @record_class.find(args, chain_options)
+        @record_class.find(
+          args,
+          chain_options.merge(error_handler: chain_error_handler)
+        )
       end
 
       def find_by(params = {})
@@ -202,7 +205,7 @@ class LHS::Record
           @record_class.request(
             chain_options
               .merge(params: chain_parameters.merge(chain_pagination))
-              .merge(error_handling: chain_error_handling)
+              .merge(error_handler: chain_error_handler)
           )
         )
       end
@@ -223,7 +226,7 @@ class LHS::Record
         merge_links _links.select { |link| link.is_a? Option }
       end
 
-      def chain_error_handling
+      def chain_error_handler
         _links.select { |link| link.is_a? ErrorHandling }
       end
 
--- a/lib/lhs/concerns/record/chainable.rb
+++ b/lib/lhs/concerns/record/chainable.rb
@@ -161,11 +161,10 @@ class LHS::Record
         push ErrorHandling.new(error_class => handler)
       end
 
-      def find(args)
-        @record_class.find(
-          args,
-          chain_options.merge(error_handler: chain_error_handler)
-        )
+      def find(*args)
+        options = chain_options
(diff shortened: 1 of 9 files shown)


cedricwider:
As soon as rubocop is happy, so am I 👍

lib/lhs/concerns/record/chainable.rb
```diff
@@ -161,8 +161,10 @@ def handle(error_class, handler)
         push ErrorHandling.new(error_class => handler)
       end
 
-      def find(args)
-        @record_class.find(args, chain_options)
+      def find(*args)
+        options = chain_options
+        options = options.merge(error_handler: chain_error_handler) if chain_error_handler.any?
```
ryabrody: Why do you not use `merge!`?
```
options.merge!(error_handler: chain_error_handler)
```
Instead of
```
options = options.merge(error_handler: chain_error_handler)
```

skaestle:
so something like

```
data = Record
  .handle(LHC::Unauthorized, ->(response) { Record.new(name: 'unknown') })
  .handle(LHC::NotFound, ->(response) { LHC::NotFound })
  .find(1, 2, 3)
data[1].name # 'unknown'
data[2] # LHC::NotFound
```

would be possible?

lib/lhs/concerns/record/request.rb
```diff
@@ -217,13 +218,31 @@ def options_for_nested_items(data, key = nil)
...
       end
 
+      # LHC supports only one error handler, merge all error handlers to one
+      # and reraise
+      def merge_error_handlers(handlers)
+        lambda do |response|
+          return_data = nil
+          error_class = LHC::Error.find(response)
+          error = error_class.new(error_class, response)
+          handlers = (handlers || []).select { |error_handler| error.is_a? error_handler.class }
```
skaestle: i would: `handlers.to_a` instead of `(handlers || [])`

pape-io:
@skaestle would be possible 👍

skaestle:
👍

lib/lhs/concerns/record/chainable.rb
pape-io: You should try to avoid `merge!` whenever possible as you are manipulating the callers original (reference) to an object, and all the origins nested objects.

http://blog.gungfu.de/archives/2005/07/05/der-unterschied-zwischen-call-by-value-und-call-by-reference/

pape-io: http://rubylearning.com/satishtalim/mutable_and_immutable_objects.html
>>>
Immutable objects make it easier to implement encapsulation. If part of an object's state is stored in an immutable object, then accessor methods can return that object to outside callers, without fear that those callers can change the object's state

lib/lhs/concerns/record/request.rb
pape-io: 👍  ill do

ryabrody:
👍 for my cmnt ☺️

PR #152, opened 2016-09-08
Inherit endpoints

pape-io:
You can inherit from previously defined records and also inherit endpoints that way:

```
class Base < LHS::Record
  endpoint 'records/:id'
end

class Example < Base
end

Example.find(1) # GET records/1
```


Averethel:
👍

fpellanda:
@pape-io Without knowing the use-case for this I would expect something like
```
Example.find(1) # GET records/1?type=examples
```

cedricwider:
👍

spec/record/endpoint_inheritance_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  context 'inherit endpoints' do
+
+    before(:each) do
+      class Base < LHS::Record
+        endpoint 'records/:id'
+      end
+
+      class Example < Base
+      end
+    end
```
fpellanda: Hm, what is the expected behaviour of this?
```
class Example < Base
  endpoint '....'
end
```

Averethel: @fpellanda I would expect that an endpoint from `Base` gets overwritten iff it's the same type, otherwise it get's added to the list.

pape-io: @Averethel what's the same type in this context ?

I would suggest the following (and I think that's also the current behaviour right now) @fpellanda :

```ruby
class Base < LHS::Record
  endpoint 'records'
end

class MegaRecords < Base
  endpoint 'records'
end

Throws clashing endpoints error
(See the last part of: https://github.com/local-ch/lhs#endpoints)
```

If they are not clashing it gets simply added to the list of `MegaRecords`.

What do you guys think? @Averethel @fpellanda 

I will add a test as soon as we agree on something.

fpellanda:
:+1: 

Maybe a default_parameter for inherited classes would be nice. But that's another feature. :)

spec/record/endpoint_inheritance_spec.rb
Averethel: I would expect the following:

```ruby
class Base < LHS::Record
  endpoint 'records'
end

class MegaRecord < Base
  endpoint 'records/mega'
end
```

```ruby
MegaRecord.find(id: 1) # GET /records/mega/1
MegaRecord.where(color: :blue) # GET /records/mega?color=blue
```

```ruby
class Base < LHS::Record
  endpoint 'records'
end

class MegaRecord < Base
  endpoint 'records/mega/:id'
end
```

```ruby
MegaRecord.find(id: 1) # GET /records/mega/1
MegaRecord.where(color: :blue) # GET /records?color=blue
```

The clashing endpoint should stay of course.

pape-io: @Averethel:
```ruby
class Base < LHS::Record
  endpoint 'records'
end

class MegaRecord < Base
  endpoint 'records/mega'
end
```
is already clashing.

If you don't use Base's base endpoint, don't use inheritance!

The second example is valid:
```ruby
class Base < LHS::Record
  endpoint 'records'
end

class MegaRecord < Base
  endpoint 'records/mega/:id'
end
```

Averethel: I would argue here. You still might have some code you'd like to share with the subclass...

Would it be hard to overwrite the endpoint in that case? If yes then let's go for clashing and review it once it's needed...

pape-io: @Averethel Would be it's own feature. You would need to implement weighted/prioritized list of endpoints.

pape-io: So far it was always clashing. As it's also more comprehensible which endpoint is taken.

PR #157, opened 2016-09-13
Item#errors.add

pape-io:
In case you want to add custom validation errors to an instance of LHS::Record:

```ruby
user.errors.add(:name, 'The name you provided is not valid.')
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,13 +539,7 @@ end
 In case you want to add custom validation errors to an instance of LHS::Record, the interface is the same then the ActiveModel::Validation one:
 
 ```ruby
-
-if name_invalid?
-
-  user.errors.add(:name, 'The name you provided is not valid.')
-
-end
-
+user.errors.add(:name, 'The name you provided is not valid.')
 ```
 
 ## Pagination
--- a/README.md
+++ b/README.md
@@ -534,6 +534,19 @@ class User < LHS::Record
 end
 ```
 
+## Custom validation errors
+
+In case you want to add custom validation errors to an instance of LHS::Record, the interface is the same then the ActiveModel::Validation one:
+
+```ruby
+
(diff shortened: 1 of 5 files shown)


lib/lhs/errors.rb
```diff
@@ -4,10 +4,10 @@ class LHS::Errors
 
   attr_reader :messages, :message, :raw
 
-  def initialize(response)
+  def initialize(response = nil)
     @messages = messages_from_response(response)
     @message = message_from_response(response)
-    @raw = response.body
+    @raw = response.body if response
   rescue JSON::ParserError # rubocop:disable Lint/HandleExceptions
```
jwagner: Shouldn't that be done in the individual methods that do the JSON.parse?

README.md
```diff
@@ -534,6 +534,13 @@ class User < LHS::Record
 end
 ```
 
+## Custom validation errors
+
+In case you want to add custom validation errors to an instance of LHS::Record, the interface is the same as ActiveModel::Validation:
```
jwagner: I'm not sure if the last claim holds: http://api.rubyonrails.org/classes/ActiveModel/Errors.html seems to have a bigger API surface than LHC::Errors.

jwagner:
Two points to consider apart from that 👍

lib/lhs/errors.rb
pape-io: @jwagner That's not part of this change. Feel free to file an issue.

README.md
pape-io: True, will drop that claim.

pape-io:
@jwagner Solved 1, the other one was not part of this PR. But I will keep it in mind for future consideration.

ryabrody:
👍

PR #167, opened 2016-10-28
! merge endpoint options when saving

pape-io:
Patch Version

## Expected Behaviour
```ruby
class RecordWithOptions < LHS::Record
  endpoint 'http://datastore/records', headers: { 'X-Header' => 'VALUE' }
end

RecordWithOptions.save(some: 'thing')
```
It should merge endpoint headers to the request done saving the record.

## Current Behaviour
It does not merge any endpoint options at all!


ryabrody:
👍

PR #169, opened 2016-11-09
Simpler complex

Averethel:
Refactored merging algorithm as discussed face to face.

It's now turning every atom into complex slowly merging them into bigger pieces. It still keeps the order.


lib/lhs/complex.rb
```diff
@@ -3,178 +3,152 @@
...
 
-  def self.reduce!(array)
-    array.flatten!
-    array.uniq!
-    return array.first if array.length == 1
-    reduce_symbols!(array)
-    return nil if array.blank?
-    array
+  def self.merge(data)
+    new(data).raw_data
```
pape-io:
```
new.reduce!(data).unwrap
```

lib/lhs/complex.rb
```diff
@@ -3,178 +3,152 @@
...
   end
 
-  def self.reduce!(array)
-    array.flatten!
-    array.uniq!
-    return array.first if array.length == 1
-    reduce_symbols!(array)
-    return nil if array.blank?
-    array
+  def self.merge(data)
```
pape-io: `reduce(data)`

lib/lhs/complex.rb
pape-io: `new.reduce!(data).unwrap`

PR #170, opened 2016-11-10
There are errors

pape-io:
PATCH VERSION

Context: validating business objects (create, update)

## Before:

In case of a custom structure content in the response body of a error response, lhs was not storing any error messages in errors. So methods like `save`, `update` etc. have been returning the boolean indicating if save or updated succeeded, but asking the `errors` object on the specific instance was telling us the opposite: no error.

This could lead to inconsistent states.

## Now:

In case of any LHC::Error the `errors` object returns and informs about errors.

### DIFF
--- a/lib/lhs/errors.rb
+++ b/lib/lhs/errors.rb
@@ -8,7 +8,8 @@ class LHS::Errors
     @messages = messages_from_response(response)
     @message = message_from_response(response)
     @raw = response.body if response
-  rescue JSON::ParserError # rubocop:disable Lint/HandleExceptions
+  rescue JSON::ParserError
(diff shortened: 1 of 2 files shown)


pape-io:
@Averethel what?!?! 🤔

Averethel:
@pape-io You just check that the error object is set. If I understand the intention correctly you'd want to forward the error message (e.g. validation error) you got in the response. Specs provided here do not do that and I think it would be worth to also test this behaviour. Unless this is already tested elsewhere.

pape-io:
@Averethel after talking in person. I'm going to add some test for the fallback structure provided through the errors object. Thank you 👍

PR #171, opened 2016-11-15
~ new kind validation configuration syntax

pape-io:
MAJOR RELEASE

## BEFOR

LHS was just supporting validations with adding `?persist=false`, `?publish=false` or `?<custom>=false` to perform validation requests.
```ruby
endpoint ':service/v2/users', validates: true
```

## NOW
We have the first endpoints that actually implement validation actions like `users/validate`. That does not work with the previous configuration syntax. So it changed:

```ruby
  endpoint ':service/v2/users', validates: { params: { persist: false } }  # will add ?persist=false to the request
  endpoint ':service/v2/users', validates: { params: { publish: false } }  # will add ?publish=false to the request
  endpoint ':service/v2/users', validates: { params: { validates: true } } # will add ?validates=true to the request
  endpoint ':service/v2/users', validates: { path: 'validate' }            # will perform a validation via :service/v2/users/validate
```


lib/lhs/concerns/item/validation.rb
```diff
@@ -8,32 +8,47 @@ module Validation
...
-            headers: { 'Content-Type' => 'application/json' }
-          )
-        )
-        true
-      rescue LHC::Error => e
-        self.errors = LHS::Errors.new(e.response)
-        false
-      end
+      raise 'No endpoint found to perform validations: https://github.com/local-ch/lhs#validation' unless validation_endpoint
+      endpoint = validation_endpoint
```
skaestle: just a small thing: store endpoint before raising, then you can reuse

README.md
```diff
@@ -551,12 +551,13 @@ unless user.valid?
...
+The parameters passed to the `validates` endpoint option are used to perform the validation:
 
 ```ruby
-class User < LHS::Record
-  endpoint ':service/v2/users', validates: 'publish'
-end
+  endpoint ':service/v2/users', validates: { persist: false }  # will add ?persist=false to the request
+  endpoint ':service/v2/users', validates: { publish: false }  # will add ?publish=false to the request
+  endpoint ':service/v2/users', validates: { validates: true } # will add ?validates=true to the request
+  endpoint ':service/v2/users', validates: '/validate'         # will perform a validation via :service/v2/users/validate
```
Averethel: Shouldn't this be something like `validates: { action: '/validate' }`? If I understand correctly now I cannot support custom endpoint & custom parameter at once.

eduardschaeli:
Regarding the interface, isn't it a bit strange that the `validates:` parameter can take a hash or a string? Wouldn't `validates: { path: '/validate'}` be more consistent?

README.md
pape-io: ?!?! what if you want to validate with `?action=/validate` would conflict. @Averethel

pape-io: BTW: Action is not a widely common term here.

pape-io:
@eduardschaeli How is path: "/validate" more comprehensible? Path's don't start with '/'

pape-io:
@Averethel @eduardschaeli you guys would prefer `{ path: 'validate' }` ?

pape-io:
@Averethel @eduardschaeli me personally I find `validates: '/validate'` more intuitive. But let's see if some body else has an opinion on that.

README.md
Averethel: @pape-io Wording / naming is something we can discuss, `path` (with omission of leading slash) works as well.
My concern is: with the current implementation when I want to use custom path I cannot also provide a custom param & vice versa.

Averethel: Also, very stupid question why not just:

`endpoint ':service/v2/users/validate', validates: true` when you want to have a custom validation action?

This way:
- we can give additional options if needed
- we can give a custom path if needed
- we avoid a lot of confusion

Or would there be clashing endpoints problem with
` endpoint ':service/v2/users'` @pape-io ?

eduardschaeli:
I'm not to concerned on the actual key name, more about having a hash or a string, not both.

pape-io:
@Averethel @eduardschaeli but to be consistent we name it 
```ruby
{ path: 'validate' } # users/validate
{ params: {persist: false} } # users?persist=false
```
Otherwise it would be less symmetrical, no?

Averethel:
@pape-io it looks reasonable enough. Have you checked my earlier suggestion, though: https://github.com/local-ch/lhs/pull/171#discussion_r88013811 ? It seems to be just simpler™️

README.md
pape-io: @Averethel clashing endpoint problem. But more over the get paremter one will get more complex

```
endpoint ':service/v2/users?persist=false', validates: true
```

Because you have to merge the get parameters you might also have to add to make the request....

pape-io: And the get parameter one is the one that is used more right now.

PR #172, opened 2016-11-18
Prevent NoMethodError in Case Item is nil

cedricwider:
Prevent NoMethodError in case item is nil


spape:
If there would be a test, I could merge it. Without I will close it in some days/weeks. 
If the problem will come up in the future again, we can easily resurrect this PR.

PR #173, opened 2016-11-21
Create through sub resources

spape:
## Create records through association

```ruby
  class Review < LHS::Record
    endpoint ':service/reviews'
  end 

  class Comment < LHS::Record
    endpoint ':service/reviews/:review_id/comments'
  end
```

### Item
```ruby
  review = Review.find(1)

  # Review#1
  # :href => ':service/reviews/1
  # :text => 'Simply awesome'
  # :comment => { :href => ':service/reviews/1/comments }

  review.comment.create(text: 'Thank you!')

  # Comment#1
  # :href => ':service/reviews/1/comments
  # :text => 'Thank you!'

  # Review#1
  # :href => ':service/reviews/1
  # :text => 'Simply awesome'
  # :comment => { :href => ':service/reviews/1/comments, :text => 'Thank you!' }
```

If the item already exists `ArgumentError` is raised.

### Expanded collection
```ruby
  review = Review.includes(:comments).find(1)

  # Review#1
  # :href => ':service/reviews/1'
  # :text => 'Simply awesome'
  # :comments => { :href => ':service/reviews/1/comments, :items => [] }

  review.comments.create(text: 'Thank you!')

  # Comment#1
  # :href => ':service/reviews/1/comments/1'
  # :text => 'Thank you!'

  # Review#1
  # :href => ':service/reviews/1'
  # :text => 'Simply awesome'
  # :comments => { :href => ':service/reviews/1/comments, :items => [{ :href => ':service/reviews/1/comments/1', :text => 'Thank you!' }] }
```

### Not expanded collection
```ruby
  review = Review.find(1)

  # Review#1
  # :href => ':service/reviews/1'
  # :text => 'Simply awesome'
  # :comments => { :href => ':service/reviews/1/comments' }

  review.comments.create(text: 'Thank you!')

  # Comment#1
  # :href => ':service/reviews/1/comments/1'
  # :text => 'Thank you!'

  # Review#1
  # :href => ':service/reviews/1
  # :text => 'Simply awesome'
  # :comments => { :href => ':service/reviews/1/comments', :items => [{ :href => ':service/reviews/1/comments/1', :text => 'Thank you!' }] }
```


lib/lhs/collection.rb
```diff
@@ -28,6 +29,10 @@ def collection?
     true
   end
 
+  def item?
+    false
```
spape: symmetrie 👍

lib/lhs/concerns/collection/internal_collection.rb
```diff
@@ -12,7 +12,7 @@ class Collection
       include Enumerable
 
       attr_accessor :raw
-      delegate :length, :size, :last, :sample, :[], :present?, :blank?, :empty?, to: :raw
+      delegate :length, :size, :last, :sample, :[], :present?, :blank?, :empty?, :<<, to: :raw
```
spape: Can you delegate `push` too? Just for completeness?

lib/lhs/concerns/proxy/create.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    private
+
+    def record_creation!
+      raise(ArgumentError, 'Record already exists') if _raw.keys != [:href] && item?
+
+      record = yield
+      merge_record_data(record)
+      reload!
```
spape: why reloading here?

lib/lhs/concerns/proxy/create.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def options_for_creation(options)
+      return options if params_from_link.blank?
+      options = {} if options.blank?
+      options.deep_merge(params: params_from_link)
+    end
+
+    def merge_record_data(record)
+      if collection?
+        _collection << record._data
```
spape: I think you should push `record._raw`

lib/lhs/concerns/proxy/create.rb
Averethel: To handle unexpanded collection

spape: Can you add this as comment?

spape: So when it's an unexpanded collection (actually it's an item then)
```ruby
# Item 
# { href: '.../reviews' }
```
then you create a nested resource
then it merges the record with the item and finally it `reloads!` which makes the item become a collection?

spape: But the problem here is that when the nested resource gets created, usually you get back the href for that nested resource:
```ruby
# { href: '.../reviews/1' }
```
when you merge this back into the unexpanded collection, you will replace it by the created item...
```ruby
# { href: '.../reviews' } <- { href: '.../reviews/1' }
# { href: '.../reviews/1' }
```

Averethel: Yes, precisely

lib/lhs/proxy.rb
```diff
@@ -15,14 +15,16 @@ def initialize(data)
...
+    reload!(options)
   end
 
-  def reload!
+  def reload!(options = nil)
     raise 'No href found' unless _data.href
-    data = _data.class.request(url: _data.href, method: :get)
+    options = {} if options.blank?
+
+    data = _data.class.request(options.merge(url: _data.href, method: :get))
```
Averethel: I thought about it for a while...
My reasoning to do it this way was that url & method seem like fixed options to properly reload the record.
Do we really want to allow overwriting here? Wouldn't that create potential security risks? (i.e. I load something completely different from what I am supposed to because I overwrite the url)

What do you think?

lib/lhs/concerns/proxy/create.rb
```diff
@@ -0,0 +1,43 @@
+require 'active_support'
+
+class LHS::Proxy
+
+  module Create
+    extend ActiveSupport::Concern
+
+    def create(data = {}, options = nil)
+      record_creation!(options) do
```
spape: Actually we forgot here to catch validation errors and provide them through errors. See 
https://github.com/local-ch/lhs/blob/3cb57d90f110acb92dc8ab9f451eca49e10293a5/lib/lhs/concerns/record/create.rb#L9
https://github.com/local-ch/lhs/blob/3cb57d90f110acb92dc8ab9f451eca49e10293a5/lib/lhs/concerns/item/save.rb#L9-L35

Averethel: good point! I'll look into that as I'm dealing with validation now anyway

Averethel: Yeah, I'm working on it right now.

lib/lhs/proxy.rb
spape: We shouldn't assume users are stupid. The opposite. When I know what I'm doing, I want to be able to do what I think is right.

spape: Actually there is an edge case...

spape: e.g.
```ruby
review.comments.create({ text: 'Food was awesome!' }, { method: :put })
```

This would be forwarded to the reload ;( 

So let's stick with the current implementation. Enforcing url, and method 👍

Averethel: I agree with that. It's just in this case this seems scary... What if I change the method to `delete` and the url to `very important parent record`?
Correct me if I'm wrong but I have a feeling that a framework should protect me from blowing up my database in such case...

Averethel:
@spape validation errors are already set correctly. I added a spec for that in ede3cfb.
Also behaviour of `create` is currently correct. It goes through build cycle (`new`) because `new` is used in `create`:
https://github.com/local-ch/lhs/blob/3cb57d90f110acb92dc8ab9f451eca49e10293a5/lib/lhs/concerns/record/create.rb#L10
and nested `create` uses `record` implementation of `create`:
https://github.com/local-ch/lhs/pull/173/files#diff-ec221a00538bb4b8b0d7bf9807ce19b4R10
(`record_from_link` returns a record class), so all is good without even touching it ;)

lib/lhs/proxy.rb
eduardschaeli: Regarding the DB, yes it absolutely does. Query planner/analyzer which every modern database has since at least 1998.

lib/lhs/concerns/proxy/link.rb
```diff
@@ -0,0 +1,24 @@
...
+    def record_from_link
+      LHS::Record.for_url(_data.href)
+    end
+
+    def endpoint_from_link
+      LHS::Endpoint.for_url(_data.href)
+    end
+
+    def params_from_link
+      return {} if !_data.href || !endpoint_from_link
```
eduardschaeli: Mhhh... but since the rest is written like this, no point in complaining :)

lib/lhs/proxy.rb
spape: Yeah. But if I tell the DB to delete a request, it will delete it  🙈  and not give me back the row without having it deleted.

PR #176, opened 2016-12-09
! a more complex includes and references case

spape:
PATCH VERSION

## BEFORE
A specific, complex `includes` and `references` case failed. One we had in SB:

```ruby
Place
  .options(auth: { bearer: token })
  .includes(contracts: :product)
  .references(contracts: { auth: { bearer: token }, product: { auth: { bearer: token } } })
  .find_by(id: online_entry_identifier)
```

Load place with auth token. Include contracts linked from place and the product, linked from any contract within place. Use auth token requesting place, contracts, and products.

Referencing deeper than one level failed (as you can see from the patch).

## NOW
Referencing is forwarded properly, so we can authenticate requests made to data included over multiple levels.

### DIFF
--- a/spec/record/includes_spec.rb
+++ b/spec/record/includes_spec.rb
@@ -407,4 +407,46 @@ describe LHS::Record do
       expect(customers.first.places.first.id).to eq 1
     end
   end
+
+  context 'more complex examples' do
+    before(:each) do
+      class Place < LHS::Record
+        endpoint 'http://datastore/places/:id'
+      end
+    end
+
+    it 'forwards complex references' do
+      stub_request(:get, "http://datastore/places/123?limit=1")
+        .to_return(body: {
+          'contracts' => {
+            'href' => "http://datastore/places/123/contracts"
+          }
+        }.to_json)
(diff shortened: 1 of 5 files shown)


spec/record/includes_spec.rb
```diff
@@ -407,4 +407,46 @@ class Place < LHS::Record
...
+            { product: { 'href' => "http://datastore/products/llo" } }
+          ]
+        }.to_json)
+      stub_request(:get, "http://datastore/products/llo?forwarded_params=for_product")
+        .to_return(body: {
+          'href' => "http://datastore/products/llo",
+          'name' => 'Local Logo'
+        }.to_json)
+      place = Place
+        .options(forward_params: 'for_place')
```
cedricwider: I don't see 'for_place' used in the stubs. Is it necessary?

spape: True. I will add it, assert it too.

spape: Added. Unveiled another bug. Fixed.

PR #177, opened 2016-12-09
fix expanding empty arrays

spape:
*PATCH VERSION*

## BEFORE

LHS was not expanding empty arrays. Wich turns out to be nil when accessing the nested data, instead of being an empty array, which would be actually more concrete.

```ruby
place = Place.includes(:contracts) # requesting contracts => { items: [] }
place.contracts # nil
```

## NOW

```ruby
place.contracts
# Data of Place#70252363673400
# :href => "http://datastore/places/123/contracts"
# :items => []
```

### DIFF
--- a/spec/record/includes_spec.rb
+++ b/spec/record/includes_spec.rb
@@ -462,7 +462,10 @@ describe LHS::Record do
           items: []
         }.to_json)
       place = Place.includes(:contracts).find('123')
-      expect(place.contracts._raw[:items]).to eq []
+      expect(
+        place.contracts.as_json
+      ).to eq('href' => 'http://datastore/places/123/contracts', 'items' => [])
+      expect(place.contracts.to_a).to eq([])
     end
   end
 end
--- a/spec/record/includes_spec.rb
+++ b/spec/record/includes_spec.rb
@@ -462,6 +462,7 @@ describe LHS::Record do
           items: []
         }.to_json)
       place = Place.includes(:contracts).find('123')
+      expect(place.contracts.collection?).to eq true
       expect(
         place.contracts.as_json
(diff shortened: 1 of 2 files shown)


spec/record/includes_spec.rb
```diff
@@ -448,5 +448,21 @@ class Place < LHS::Record
...
+            'href' => "http://datastore/places/123/contracts"
+          }
+        }.to_json)
+      stub_request(:get, "http://datastore/places/123/contracts")
+        .to_return(body: {
+          href: "http://datastore/places/123/contracts",
+          items: []
+        }.to_json)
+      place = Place.includes(:contracts).find('123')
+      expect(place.contracts._raw[:items]).to eq []
```
ryabrody: Why are you doing the expect on the `_raw[:items]` method instead of `contracts` as you said in the description of the PR like this:
```
expect(place.contracts).to eq []
```

spape: because `place.contracts` is actually an LHS data object:
```ruby
Data of Place#70252363673400
:href => "http://datastore/places/123/contracts"
:items => []
```

spape: Do you think, that this: 
https://github.com/local-ch/lhs/blob/d42fc01814b9713a6486e80581ed7987eadfda28/spec/record/includes_spec.rb#L465-L469
would be more comprehensible?

spape:
```ruby
expect(place.contracts.collection?).to eq true
      expect(
        place.contracts.as_json
      ).to eq('href' => 'http://datastore/places/123/contracts', 'items' => [])
      expect(place.contracts.to_a).to eq([])
```

PR #178, opened 2016-12-09
Improve inspect

spape:
*MINOR VERSION*

When inspecting Records or Data of Records, identifying which objects you deal with and determine where your are currently located in a nested data structure was difficult.

This improvements helps identifying inspected Records, by displaying either href, the id attribute or the object_id in this order.

Further more, this change shows you the current location inside nested data structures, so you now where are you currently inspecting:

## BEFORE

```ruby
place = Place.find('INHJ_EoVxQgsHk65RkDUvQ')
 => Place#70235251510900
:href => "http://datastore-stg.lb-service.sunrise.intra.local.ch/v2/places/INHJ_EoVxQgsHk65RkDUvQ"
:contacts => [{:type=>"phone", :details=>"041 360 45 50", :refuse_advertising=>false, :preferred=>false}]

place.contacts.first
 => Data of Place#70235251323100
:type => "phone"
:details => "041 360 45 50"
:refuse_advertising => false
:preferred => false 
```

## NOW

```ruby
place = Place.find('INHJ_EoVxQgsHk65RkDUvQ')
=> Place http://datastore-stg.lb-service.sunrise.intra.local.ch/v2/places/INHJ_EoVxQgsHk65RkDUvQ
:href => "http://datastore-stg.lb-service.sunrise.intra.local.ch/v2/places/INHJ_EoVxQgsHk65RkDUvQ"
:contacts => [{:type=>"phone", :details=>"041 360 45 50", :refuse_advertising=>false, :preferred=>false}]

place.contacts.first
=> Data of Place http://datastore-stg.lb-service.sunrise.intra.local.ch/v2/places/INHJ_EoVxQgsHk65RkDUvQ
> contacts > 0
:type => "phone"
:details => "041 360 45 50"
:refuse_advertising => false
:preferred => false 
```


lib/lhs/concerns/inspect.rb
```diff
@@ -5,13 +5,46 @@ module Inspect
...
   private
 
+  def _inspect_id
+    _root.href || (_root.item? ? _root.id : nil) || object_id
+  end
+
+  def _inspect_path
+    current = self
+    path = []
+    while current.parent
```
skaestle: looks like something that could be handled recursively? maybe worth investing some time

Averethel: 👍  @skaestle, let's try recursion

lib/lhs/concerns/inspect.rb
```diff
@@ -5,13 +5,46 @@ module Inspect
...
+        _inspect_path
+      ].compact.join(' '),
       pretty_raw
     ].compact.join("\n")
   end
 
   private
 
+  def _inspect_id
+    _root.href || (_root.item? ? _root.id : nil) || object_id
```
Averethel: Is `href` really a good solution here? If so maybe we want a line break after it? Otherwise one needs to scroll to see the path.

lib/lhs/concerns/inspect.rb
```diff
@@ -5,13 +5,49 @@ module Inspect
...
+    if parent_raw.is_a?(Array)
+      parent_raw.each_with_index do |element, index|
+        path.push(index) if element == current._raw
+      end
+    elsif parent_raw.is_a?(Hash)
+      parent_raw.each do |key, value|
+        path.push(key) if value == current._raw
+      end
+    end
+    _collect_parents_for_inspect!(path, current.parent) if current.parent
```
Averethel: The `if` is not needed. `current.parent` is always truthy at this point 'cause of line 37.

lib/lhs/concerns/inspect.rb
spape: path move to new line now 👍

PR #183, opened 2016-12-21
Load and merge created data through location header

spape:
*PATCH VERSION*

When creating records, location header has not been considered when creation endpoint returns no content.

## BEFORE

```ruby
contact_person = ContactPerson.create(first_name: 'Sebastian')
# POST http://datastore/contact_persons
# RETURNS 204 body: '', headers: { Location: 'http://datastore/contact_persons/1' }
contact_person.href # nil
```

## NOW

Location-Header is considered, fetched and merged with raw data if creation endpoint provides such header on creation.

```ruby
contact_person = ContactPerson.create(first_name: 'Sebastian')
contact_person.href # http://datastore/contact_persons/1
```

### DIFF
--- a/lib/lhs/concerns/item/save.rb
+++ b/lib/lhs/concerns/item/save.rb
@@ -15,29 +15,34 @@ class LHS::Item < LHS::Proxy
     def save!(options = {})
       options = options.present? ? options.dup : {}
       data = _data._raw.dup
-      if href.present?
-        url = href
-      else
-        endpoint = endpoint_for_persistance(data, options)
-        url = url_for_persistance(endpoint, data, options)
-        endpoint.remove_interpolated_params!(data)
-        endpoint.remove_interpolated_params!(options.fetch(:params, {}))
-        options.merge!(endpoint.options.merge(options)) if endpoint.options
-      end
-
-      options = options.merge(method: :post, url: url, body: data.to_json)
-      options[:headers] ||= {}
-      options[:headers].merge!('Content-Type' => 'application/json')
-
-      data = record_for_persistance.request(options)
-      _data.merge_raw!(data)
-      true
+      url = url_for_persistance!(options, data)
+      create_and_merge_data!(
+        apply_default_creation_options(options, url, data)
+      )
     rescue LHC::Error => e
       self.errors = LHS::Errors.new(e.response)
       raise e
     end
 
     private
+
+    def apply_default_creation_options(options, url, data)
+      options = options.merge(method: :post, url: url, body: data.to_json)
+      options[:headers] ||= {}
+      options[:headers].merge!('Content-Type' => 'application/json')
+      options
+    end
+
+    def create_and_merge_data!(options)
+      direct_response_data = record_for_persistance.request(options)
+      _data.merge_raw!(direct_response_data)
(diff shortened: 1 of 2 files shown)


lib/lhs/concerns/item/save.rb
```diff
@@ -56,10 +61,19 @@ def record_for_persistance
       _data.class
     end
 
-    def url_for_persistance(endpoint, data, options)
-      endpoint.compile(
-        merge_data_with_options(data, options)
-      )
+    def url_for_persistance!(options, data)
+      if href.present?
```
Averethel: How about a guard clause?
`return href if href.present?`

lib/lhs/concerns/item/save.rb
```diff
@@ -56,10 +61,19 @@ def record_for_persistance
...
+        href
+      else
+        endpoint = endpoint_for_persistance(data, options)
+        url = endpoint.compile(
```
Averethel: I'm a bit confused here. How are changes to `endpoint` (L72 & L73) affecting the url. Shouldn't the endpoint be compiled to url at the very end?

Averethel: Also maybe it's a good idea to consider using `tap` here?

spape: L72, L73 do not affect the URL, but the sources, used to compile the url. If `{ feedback_id: 123 }` was provided through params and it got backed into the url: `http://datastore/feedbacks/1` it has to be removed from the params, especially when they are send as body (in case of creation/post).

spape: And url is compiled at the very end, but you can only clean up the sources after it got compiled, otherwise you cant compile anymore.

spape: If you make an implementation example for tap. I might consider it.

Averethel: So url compilation "pollutes" the enpoint object with some data? Do I get this correctly?
Or is this compilation lazy and it's only resolved when url variable is actually accessed? And some preprocessing needs to be done before?
In both cases maybe we can simplify the compilation process a bit (out of scope here, but something to consider int the future™️)

Averethel: and here goes `tap` example:
```ruby
        endpoint = endpoint_for_persistance(data, options)
        endpoint.compile(
          merge_data_with_options(data, options)
        ).tap do
          endpoint.remove_interpolated_params!(data)
          endpoint.remove_interpolated_params!(options.fetch(:params, {}))
          options.merge!(endpoint.options.merge(options)) if endpoint.options
        end
```

Averethel:
@spape thanks for the explanation. I think I get it now.

lib/lhs/concerns/item/save.rb
spape: No the endpoint object is not getting polluted. The endpoint is the only object that can clean sources required to get and endpoint like `params: {}` or a `body: {}`. It simply removes parameters that are moved to the url during compilation: 
1. `params: { feedback_id: 1 }` 
2. -(compile)> 'http://datastore/feedbacks/1'
3. `{ feedback_id: }` not needed in params anymore
4. `params: {}`

PR #186, opened 2017-01-13
Add proper autoloading with middleware fixes #185

### LINKED ISSUE #185: Preload LHS classes breaks rails's model autoloading
LHS should not require classes on server startup when ```config.cache_classes = false```

```ruby 
# lib/lhs.rb
# Preload all the LHS::Records that are defined in app/models
class Engine < Rails::Engine
  initializer 'Load all LHS::Records from app/models/**' do |app|
    Dir.glob(app.root.join('app/models/**/*.rb')).each do |file|
      require file if File.read(file).match('LHS::Record')
    end
  end
end
```

If this is really required, load these classes before each request when ```config.cache_classes = false```

fpellanda:


spape:
Could you explain the difference to before? Now instead of loading all LHS::Records into memory on app startup. You load it when the first Rack request comes in ?!?!

spape:
Rails' class_caching doesn't work with LHS for the following reason.

In Rails, when a class is called for the first time, like `User`, it loads the file via naming conventions and caches it. 

There is an edge case in LHS, when you request objects via an LHS::Record, like `Customer` it might be able to require the file and load this class into memory to perform the request, but, the data returned, might include links to other records. As we cast to those records when navigating those links, via data traversing, or using Record options from other Records when performing things like include, those classes have to be around when evaluating those links and looking up possible matched Records.

Examples for Record lookups via `href`: 
https://github.com/local-ch/lhs/blob/includes-all/lib/lhs/concerns/record/request.rb#L255
https://github.com/local-ch/lhs/blob/includes-all/lib/lhs/item.rb#L75

Could you elaborate on how the change improves the current situation? We might find another way, or load Records when they are needed. 

I also think adding middleware compared to App-Startup initialized adds way more overhead.

spape:
https://github.com/local-ch/lhs/issues/185

fpellanda:
Yes, load all files before each request with ```require_dependency``` if cache classes is false. ```require_dependency``` does not load the file again if it's already loaded, but reloads it when it changed.

This has the following advantages:
- You don't have unexpected class loading during startup
- When you add a file with LHS::Record, the file will be immediately available for LHS without restart
- When you change a file that was loaded with ```require_dependency``` the changes will be available immediately.
- In Production resp. when cache_classes is true, all classes will be autoloaded on startup anyway.

Related to middleware, this is only in development mode... and only like an additional method call.

@spape Do you think something will break with this PR?

spape:
You might want to merge this before this major change:
https://github.com/local-ch/lhs/pull/190

spape:
I'll make a test deployment with this (as discussed), so that I can merge/release it today.

spape:
Worked https://www--location-app--test-lhs-autoloading.play.intra.local.ch/

PR #187, opened 2017-01-25
Fix referencing options in case includes is an array of symbols

spape:
*PATCH VERSION*

## Before
It was not possible to apply the `references` options to all symbols of the `includes` array.

```ruby
options = { headers: { 'Authentication': 'Bearer 123' } }
customer = Customer
  .includes(:electronic_addresses, :contact_addresses)
  .references(
    electronic_addresses: options,
    contact_addresses: options
  )
  .find(1)
```

It was failing, because by mistake, the `references` hash was converted to an array internally, being applied only on the first symbol of the `includes` array.

## Now

The `references` hash, as expected, gets applied to all symbols in the `includes` array, according to their key: `:electronic_addresses` => `electronic_addresses: options`, etc.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,3 +27,5 @@ Style/SignalException:
 Rails/DynamicFindBy:
   Enabled: false
   
+Metrics/BlockLength:
+  Enabled: false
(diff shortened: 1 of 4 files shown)


spape:
@cedricwider please try in your branch :) and let me know if it works.

PR #188, opened 2017-01-27
+ next? and previous?

spape:
### Pagination Links

When endpoints provide indicators for current page position with links (like `next` and `previous`), LHS provides some functionalities to interact/use those links/information:

`next?` Tells you if there is a next link or not.

`previous?` Tells you if there is a previous link or not.


spec/record/pagination_links_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      Customer.where(name: name, page: page)
+    end
+
+    context 'next link is present, previous is not' do
+      let(:name) { 'Simpl' }
+      let(:page) { 2 }
+      let(:body) do
+        {
+          next: "http://datastore/customer?name=#{name}&page=3",
+          previous: "http://datastore/customer?name=#{name}&page=1",
```
crackofdusk: This does not match the context description: there is in fact a `previous` link.

spape: Fixed https://github.com/local-ch/lhs/pull/188/commits/443bbaef92807f30744b9da240eb5f19484783f2

PR #190, opened 2017-01-30
Record.includes_all(resource: :sub_resource)

spape:
*MAJOR VERSION*

## Include linked resources
  
When fetching records, you can specify in advance all the linked resources that you want to include in the results. With `includes` or `includes_all` (to enforce fetching all remote objects for paginated endpoints), LHS ensures that all matching and explicitly linked resources are loaded and merged.

### `includes_all` for paginated endpoints

In case endpoints are paginated and you are certain that you'll need all objects of a set and not only the first page/batch, use `includes_all`.

LHS will ensure that all linked resources are around by loading all pages (parallelized/performance optimized).

```ruby
customer = Customer.includes_all(contracts: :products).find(1)

# GET http://datastore/customers/1
# GET http://datastore/customers/1/contracts?limit=100
# GET http://datastore/customers/1/contracts?limit=10&offset=10
# GET http://datastore/customers/1/contracts?limit=10&offset=20
# GET http://datastore/products?limit=100
# GET http://datastore/products?limit=10&offset=10

customer.contracts.length # 33
customer.contracts.first.products.length # 22
```

## Important Facts:
- `all` and the internal mechanism that is self-paginating works with provided `options` (e.g. auth) now, was not the case before, before it was just accepting (params). This also makes this change a non-backwards-compatible change.
- `all` and the internal mechanism that is self-paginating makes parallel request to fetch a batch now (performance optimized), was not performance optimized before
- A warning is send to std_err if `includes` is used, but fetched result is paginated.


lib/lhs/concerns/record/all.rb
```diff
@@ -11,10 +11,12 @@ module ClassMethods
       # compute the amount of left over requests, do all the the left over requests
       # for the following pages and concatenate all the results in order to return
       # all the objects for a given resource.
-      def all(params = {})
-        limit = params[limit_key] || LHS::Pagination::DEFAULT_LIMIT
-        data = request(params: params.merge(limit_key => limit))
-        request_all_the_rest(data, params) if paginated?(data._raw)
+      def all(options = {})
+        options ||= {}
```
scan: Do you really need this when the default param is already set?

Averethel: @scan  `Record.all(nil)`

lib/lhs/concerns/record/all.rb
```diff
@@ -11,10 +11,12 @@ module ClassMethods
...
+        options[:params] ||= {}
+        options[:params] = options[:params].merge(limit_key => options[:params][limit_key] || LHS::Pagination::DEFAULT_LIMIT)
```
Averethel: why not `merge!`?
_minor_

lib/lhs/concerns/record/chainable.rb
```diff
@@ -190,6 +196,12 @@ def includes(*args)
         push Include.new(Chain.unfold(args))
       end
 
+      def includes_all(*args)
+        chain = push Include.new(Chain.unfold(args))
```
Averethel: parens `push(Include.new(Chain.unfold(args)))`

_minor_

lib/lhs/concerns/record/chainable.rb
```diff
@@ -265,6 +285,28 @@ def resolved_options
 
       private
 
+      def includes_all_to_references(args, parent = nil)
```
Averethel: Since we elsewhere follow pattern of splitting such methods in smaller parts based on a case how about doing this also here?

lib/lhs/pagination.rb
```diff
@@ -84,8 +84,8 @@ def current_page
     offset
   end
 
-  def next_offset
-    current_page + 1
+  def next_offset(step = 1)
```
Averethel: How about extracting those subclasses to separate files?

lib/lhs/concerns/record/all.rb
spape: Yes. @scan When nil is passed, default parameter is not applied.

lib/lhs/concerns/record/all.rb
spape: I try to prevent `merge!` as much as possible to keep hashes as immutable as possible. Otherwise those changes might bubble up.

lib/lhs/concerns/record/chainable.rb
spape: done 👍

spec/record/includes_all_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    let!(:products_request_page_3) do
+      additional_products_request(20, 2)
+    end
+
+    it 'includes all linked business objects no matter pagination' do
+      customer = Customer
+        .includes_all(contracts: :products)
+        .find(1)
+      expect(customer.contracts.length).to eq 33
```
skaestle: would be nice to have these in a `let(:expected_contracts_length)`

lib/lhs/concerns/record/all.rb
scan: Yes it can, but that would be a deliberate user error, I guess?

lib/lhs/concerns/record/chainable.rb
spape: done 👍

lib/lhs/pagination.rb
spape: done 👍

lib/lhs/concerns/record/all.rb
spape: I rather want the connivence of NOT running into an error when doing `Record.all(options)  # where options is nil` just because somebody wants to save one line of code.

PR #192, opened 2017-01-30
Accessors in collection

Averethel:
```ruby
class Record < LHS::Record
  endpoint ':datastore/records`'
end
```

Response:
```json
{
  "items":  [{ "name": "Steve" }],
  "extra":  "extra",
  "total":  1
}
```

# Before
```ruby
Record.where.extra

# => 
# undefined method `extra' for #<LHS::Collection::InternalCollection::Collection:0x007f9b59b1a7c8>
```

# After
```ruby
Record.where.extra

# => 
# "extra"
```

### DIFF
--- a/lib/lhs/concerns/proxy/accessors.rb
+++ b/lib/lhs/concerns/proxy/accessors.rb
@@ -1,0 +1,73 @@
+require 'active_support'
+
+class LHS::Proxy
+  module Accessors
+    extend ActiveSupport::Concern
+
+    private
+
+    # Keywords that would not be forwarded via method missing
+    # FIXME: Extend the set of keywords
+    BLACKLISTED_KEYWORDS = %w(new proxy_association)
+
+    def set(name, value)
+      key = name.to_s.gsub(/=$/, '')
+      _data._raw[key.to_sym] = value
+    end
+
+    def get(name, *args)
+      name = args.first if name == :[]
+      value = _data._raw[name.to_s]
+      if value.nil? && _data._raw.present?
+        value = _data._raw[name.to_sym]
+        value = _data._raw[name.to_s.classify.to_sym] if value.nil?
+      end
+      if value.is_a?(Hash)
+        handle_hash(value)
+      elsif value.is_a?(Array)
+        data = LHS::Data.new(value, _data)
+        collection = LHS::Collection.new(data)
+        LHS::Data.new(collection, _data)
+      else
+        convert(value)
+      end
+    end
+
+    def convert(value)
+      return value unless value.is_a?(String)
+      if date_time?(value)
+        Time.zone.parse(value)
+      elsif date?(value)
+        Date.parse(value)
+      else
+        value
+      end
+    end
+
+    def handle_hash(value)
+      record = LHS::Record.for_url(value[:href]) if value[:href]
+      data = LHS::Data.new(value, _data)
+      if record
+        record.new(data)
+      else
+        data
+      end
+    end
+
+    def date?(value)
+      value[date_time_regex, :date].presence
+    end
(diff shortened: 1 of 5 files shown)


lib/lhs/concerns/proxy/accessors.rb
```diff
@@ -0,0 +1,73 @@
...
+
+class LHS::Proxy
+  module Accessors
+    extend ActiveSupport::Concern
+
+    private
+
+    # Keywords that would not be forwarded via method missing
+    # FIXME: Extend the set of keywords
+    BLACKLISTED_KEYWORDS = %w(new proxy_association)
```
spape: Constants are only private if you use `private_constant`. I would put it on top. (minor)

lib/lhs/concerns/proxy/accessors.rb
```diff
@@ -0,0 +1,73 @@
...
+      name = args.first if name == :[]
+      value = _data._raw[name.to_s]
+      if value.nil? && _data._raw.present?
+        value = _data._raw[name.to_sym]
+        value = _data._raw[name.to_s.classify.to_sym] if value.nil?
+      end
+      if value.is_a?(Hash)
+        handle_hash(value)
+      elsif value.is_a?(Array)
+        data = LHS::Data.new(value, _data)
```
spape: Suggestion: Refactor this block to `handle_collection` (symmetry) and rename the above method to `handle_item` to be correct *minor*

spape: and it the new context of accessor we might give a more precise name, like `return_item`, `return_collection` or `access_item` / `access_collection`

spec/collection/accessors_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+
+describe LHS::Collection do
+  let(:datastore) { 'http://local.ch/v2' }
+  let(:items) { [{name: 'Steve'}] }
+  let(:extra) { 'extra' }
+  let(:collection) { Record.where }
```
spape: soon. `fetch`

spec/collection/accessors_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    let(:response_data) do
+      {
+        items: items,
+        extra: extra,
+        total: 1
+      }
+    end
+
+    before(:each) do
+      LHC.config.placeholder('datastore', datastore)
```
spape: Try to not copy this outdated pattern. I know there are still a lot of tests with the additional placeholder configuration. But it's just an overhead I started some time ago to continue writing tests the following way:

```ruby
before(:each) do
class Record < LHS::Record
  endpoint 'http://datastore/records'
end
stub_request(:get, %r{http://datastore/v2/records})
  .to_return(body: response_data.to_json)
end
```

spec/collection/accessors_spec.rb
spape: and BTW, this gives you the chain, not the collection ;)

spape:
Rubocop ;P

spape:
I recommend merging this before this major change: https://github.com/local-ch/lhs/pull/190

spec/collection/accessors_spec.rb
Averethel: yeah I wasn't sure which one is the current state of art so I used some other collection spec as a template.

spec/collection/accessors_spec.rb
Averethel: True but it gets resolved on access no?

Averethel:
@spape rubocop? what ruboccop? :trollface:

PR #193, opened 2017-02-03
Chainable `all`

spape:
*MAJOR VERSION*

`all` fetches all records from the service by doing multiple requests and resolving endpoint pagination if necessary.

```ruby
data = Record.all
data.count # 998
data.length # 998
```

`all` is chainable and has the same interface like `where` (See: [Find multiple records](https://github.com/local-ch/lhs#find-multiple-records))

```ruby
Record.where(color: 'blue').all
Record.all.where(color: 'blue')
Record.all(color: 'blue')
# All three are doing the same thing: fetching all records with the color 'blue' from the endpoint while resolving pagingation if endpoint is paginated
```

-----------------

- Moved `all` to Chainable
- Moved the internal logic that paginates endpoints to LHS::Record#request to make it available to all kind of requests
- Also fixes some problems with `all` in version https://github.com/local-ch/lhs/releases/tag/v8.0.0


lib/lhs/concerns/record/request.rb
```diff
@@ -160,6 +168,42 @@ def skip_loading_includes?(data, included)
         end
       end
 
+      # After fetching the first page,
+      # we can evaluate if there are further remote objects remaining
+      # and after preparing all the requests that have to be made in order to fetch all
+      # remote items during this batch, they are fetched in parallel
+      def load_and_merge_all_the_rest!(data, options)
```
cedricwider: _ALL_ the rest...? How about something like `load_and_merge_remaining_objects!`?

spape: Yeah. I think `load_and_merge_remaining_objects` is more political correct for the remaining objects ;). will do.

PR #194, opened 2017-02-03
+ how to test

spape:
## Testing: How to write tests when using LHS

[WebMock](https://github.com/bblimke/webmock)! 

Best practice is to let LHS fetch your records and Webmock to stub/mock endpoints responses. 
This follows the [Black Box Testing](https://en.wikipedia.org/wiki/Black-box_testing) approach and prevents you from building up constraints to LHS' internal structures/mechanisms, which will break when we change internal things.
LHS provides interfaces that result in HTTP requests, this is what you should test.

```ruby
let(:contracts) do
  [
    {number: '1'},
    {number: '2'},
    {number: '3'}
  ]
end

before(:each) do
  stub_request(:get, "http://datastore/user/:id/contracts")
    .to_return(
      body: {
        items: contracts,
        limit: 10,
        total: contracts.length,
        offset: 0
      }.to_json
    )
end

it 'displays contracts' do
  visit 'contracts'
  contracts.each do |contract|
    expect(page).to have_content(contract[:number])
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -844,6 +844,62 @@ end
 Example.find(1) # GET records/1
 ```
 
+## Testing: How to write tests when using LHS
+
+[WebMock](https://github.com/bblimke/webmock)! 
+
+Best practice is to let LHS fetch your records and Webmock to stub/mock endpoints responses. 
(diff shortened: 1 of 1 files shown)


skaestle:
💋

eduardschaeli:
💃

PR #197, opened 2017-02-09
Fix includes all for empty arrays

spape:
Fixes case where response contains empty array but LHS tries to access `._raw` on empty array.

### DIFF
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -175,7 +175,7 @@ class LHS::Record
       def load_and_merge_remaining_objects!(data, options)
         if paginated?(data._raw)
           load_and_merge_paginated_collection!(data, options)
-        elsif data.collection? && paginated?(data.first.try(&:_raw))
+        elsif data.collection? && paginated?(data.first.try(:_raw))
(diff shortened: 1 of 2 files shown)


lib/lhs/concerns/record/request.rb
```diff
@@ -175,7 +175,7 @@ def skip_loading_includes?(data, included)
       def load_and_merge_remaining_objects!(data, options)
         if paginated?(data._raw)
           load_and_merge_paginated_collection!(data, options)
-        elsif data.collection? && paginated?(data.first._raw)
+        elsif data.collection? && paginated?(data.first.try(&:_raw))
```
scan: Are you sure it"s not just `try(:_raw)`, or in newer ruby, just `data.first&._raw`?

spape: The same 
```ruby
# File activesupport/lib/active_support/core_ext/object/try.rb, line 28
  def try(*a, &b)
    if a.empty? && block_given?
      yield self
    else
      __send__(*a, &b)
    end
  end
```

Any way I will use `try(:_raw)` to cause less confusion. And I would like to not starting using new Ruby features like `&.` yet in this library.
👍

PR #198, opened 2017-02-20
+ test case for includes

spape:
*PATCH VERSION*

Including linked objects, when reference was an array of links was not working properly.
```ruby
# GET /place
{
  category_relations: [
    { href: 'http://datastore/category/1' }, { href: 'http://datastore/category/2' }
  ]
}
```

```ruby
Place
  .includes(:category_relations)
  .find(1, 2)
```

### DIFF
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -62,8 +62,8 @@ class LHS::Record
           item = item[i] if item.is_a? LHS::Collection
           link = item[key.to_sym]
           if link && link.collection?
-            link.each_with_index do |link, j|
-              link.merge_raw!(addition[i + j]) if link.present?
+            link.each_with_index do |item, j|
+              item.merge_raw!(addition[i + j]) if item.present?
             end
           elsif link.present?
             link.merge_raw!(addition[i])
@@ -413,8 +413,8 @@ class LHS::Record
       def url_option_for(item, key = nil)
         link = key ? item[key] : item
         if link && link.collection?
-          link.map do |link|
-            { url: link.href } if link.present? && link.href.present?
+          link.map do |item|
+            { url: item.href } if item.present? && item.href.present?
           end.compact
         elsif link.present? && link.href.present?
           { url: link.href }
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -65,8 +65,8 @@ class LHS::Record
             link.each_with_index do |link, j|
               link.merge_raw!(addition[i + j]) if link.present?
             end
-          else
-            link.merge_raw!(addition[i]) if link.present?
+          elsif link.present?
+            link.merge_raw!(addition[i])
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/lhs/concerns/record/request.rb
```diff
@@ -61,7 +61,13 @@ def extend_base_collection!(data, addition, key)
         data.each_with_index do |item, i|
           item = item[i] if item.is_a? LHS::Collection
           link = item[key.to_sym]
-          link.merge_raw!(addition[i]) if link.present?
+          if link && link.collection?
+            link.each_with_index do |link, j|
```
scan: Not a fan of shadowing outer variables with params of `do`, in this case `link`. Perhaps use a different name for the "inner" `link' here?

lib/lhs/concerns/record/request.rb
```diff
@@ -406,7 +412,13 @@ def single_request(options)
 
       def url_option_for(item, key = nil)
         link = key ? item[key] : item
-        return { url: link.href } if link.present? && link.href.present?
+        if link && link.collection?
+          link.map do |link|
```
scan: See above about shadowing variables inside `do`.

lib/lhs/concerns/record/request.rb
spape: 👍 done

lib/lhs/concerns/record/request.rb
spape: 👍 done

lib/lhs/concerns/record/request.rb
```diff
@@ -61,7 +61,13 @@ def extend_base_collection!(data, addition, key)
         data.each_with_index do |item, i|
           item = item[i] if item.is_a? LHS::Collection
           link = item[key.to_sym]
-          link.merge_raw!(addition[i]) if link.present?
+          if link && link.collection?
```
Averethel: How about a guard clause for `link.present?`
Something along:

```ruby
next if !link.present?
link.merge_raw!(addition[i]) if !link.collection?

link.each_with_index do |item, j|
  item.merge_raw!(addition[i + j]) if item.present?
end
```

I find this a bit more readable

lib/lhs/concerns/record/request.rb
```diff
@@ -406,7 +412,13 @@ def single_request(options)
 
       def url_option_for(item, key = nil)
         link = key ? item[key] : item
-        return { url: link.href } if link.present? && link.href.present?
+        if link && link.collection?
```
Averethel: I would prefer similar style as I suggested here: https://github.com/local-ch/lhs/pull/198/files#r102200590

lib/lhs/concerns/record/request.rb
spape: @Averethel you mean
```
next if !link.present?
link.merge_raw!(addition[i]) && next if !link.collection?

link.each_with_index do |item, j|
  item.merge_raw!(addition[i + j]) if item.present?
end
```

spape: ???

Averethel: yeah, I missed one `next` in the example. Good spot 👍

PR #201, opened 2017-02-27
! always load LHS classes once

Averethel:
_PATCH_

Fixes endpoint discovery in when classes were not initially loaded.


spape:
Add test, then we are fine. You can instrumentalize the dummy app, if you have to.

PR #204, opened 2017-03-03
Compatibility with ActiveModel

cedricwider:
Compatibility with ActiveModel

ActiveModel expects the errors object to always be present
!~ Introduce 'clear' on LHS::Error to mimic ActiveModel::Error behaviour

### DIFF
--- a/spec/item/validation_spec.rb
+++ b/spec/item/validation_spec.rb
@@ -95,7 +95,6 @@ describe LHS::Item do
       expect(user.valid?).to eq true
       expect(user.errors.messages).to be_empty
     end
-
   end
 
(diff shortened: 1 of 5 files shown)


spec/item/validation_spec.rb
```diff
@@ -93,7 +93,7 @@ class User < LHS::Record
       user.email = 'steve@local.ch'
       successful_validation
       expect(user.valid?).to eq true
-      expect(user.errors).to be_nil
```
spape: Suggesting major, because of that change!

cedricwider: I agree 👍

PR #206, opened 2017-03-10
Fetch record by href

scan:


spape:
*MINOR VERSION*

spape:
BTW: could you add something to the docs too?

scan:
@spape Did done so.

PR #208, opened 2017-03-15
! forwarding options to additionally expanded list of links

spape:
*PATCH VERSION*

## Background information

When crossing services/domains, in can happen that one resource contains a link to a collection that itself, just contains links e.g.:

```
GET /customers/1 (CUSTOMER MANAGEMENT)
{
  places: { 'href' : 'http://datastore/customers/1/places' }
}
```

```
GET http://datastore/customers/1/places (DATASTORE)
{
  items: [
    { 'href' : 'http://datastore/place/3' },
    { 'href' : 'http://datastore/place/5' }
  ]
}
```

When requesting LHS to `.includes` the linked items, instead of just requesting the link and merging the back the "empty" links to the original customer, it also additional expands those initially requested items with an additional request-batch.

## Previously unexpected behaviour

When using references options, in order to apply options to the requests made to fetch the linked resources:

```ruby
Customer
  .includes(:places)
  .references(places: { auth: { bearer: '123'}})
  .find(1)
```

LHS was just using those options for the initial request to fetch the list of "empty" links, but was not forwarding those options also to the additional expansion.

## Behaviour now. Fix.

Now LHS also uses the specific refence (options) when additionally expanding those empty "links".


lib/lhs/concerns/record/request.rb
```diff
@@ -140,9 +140,10 @@ def options_for_data(data, included = nil)
         url_option_for(data, included)
       end
 
-      def expand_addition!(data, included)
+      def expand_addition!(data, included, reference)
```
fpellanda: `options` instead of `reference`, or `options = {}`?

spape: Well, options would shadow the variable inside this scope, that's already called options. Any suggestions @fpellanda ?

spape: Can you elaborate on why you need a unit test? All functionalities in LHS is covered by intergration tests, this PR just implements a patch and does not implement a new method. Also, expand_addition is unlikely to be shared with/called from other methods, it's just there to split sequentiell executions in blocks that are easier to understand.

spape:
@fpellanda Can you elaborate on why you need a unit test? All functionalities in LHS is covered by intergration tests, this PR just implements a patch and does not implement a new method. Also, expand_addition is unlikely to be shared with/called from other methods, it's just there to split sequentiell executions in blocks that are easier to understand.

lib/lhs/concerns/record/request.rb
fpellanda: Hm, rename one of them? What's the difference between these two options? `new_options`, `additional_options`, `options_to_expand`?

fpellanda:
> @fpellanda Can you elaborate on why you need a unit test? All functionalities in LHS are covered by intergration tests, this PR just implements a patch and does not implement a new method. Also, expand_addition is unlikely to be shared with/called from other methods, it's just there to split sequentiell executions in blocks that are easier to understand.

For me it's not understandable, what the input and output of these methods are. Those methods are not trivial. It's quite hard to work on these method when you're not using this all the day.

Additionally, handling edge cases cannot be done in integration tests.

Who should be able to fix stuff in this library when you are away? Maybe he should say if it's enough for him. For me it wouldn't.

spape:
@fpellanda okay, will implement a unit test.

lib/lhs/concerns/record/request.rb
spape: What has `new_options`, `additional_options` and `options_to_expand` to do with the `expand_addition!` function? Are you making things up now?

spape: Well, trying to put myself into the situation where I don't know the implementation of `expand_addition!`, I would read the code as following:
>
Expand an addition, where the addition is the incoming data, navigating the provided key called `included`. Inside the method options are prepared in order to execute a request that extends the base data. Those prepared options take into account the reference that is provided from the outside.

spape:
@fpellanda beside that, the whole services, and remote object retrival topic is a complex one. That does not mean, that the method "expand_addition" is a complex method. It expands an addition. That simple.

nerdinand:
It still doesn't quite seem to be doing what it should. If I pass an additional parameter to `references`, it seems to get ignored, e.g. with:

```ruby
Customer
  .includes(:places)
  .references(places: { auth: { bearer: '123'}, limit: 1000})
  .find(1)
```

I expect `?limit=1000` in the request url, but it's not there.

lib/lhs/concerns/record/request.rb
spape: Answering "What's the difference between these two options?".
Well inside the method `expand_addition!` the local variable `options` is used to prepare request options, used to perform a request that retreives data that's used for the actual expansion. `reference` instead describes additional options, that have to be taken into account when performing requests, as references are provided by consumers, who have the need to apply additional request options to requests done by LHS.

spape:
well, because it's 
```ruby
.references(places: { auth: { bearer : '123' }, params: { limit: 1000 } })
```
otherwise, `auth` would be a query param too -.-

spape:
And BTW @fpellanda, I have to disagree with `Additionally, handling edge cases cannot be done in integration tests.`. With integration tests you can reach 100% test coverage. So also edge cases are included. As the whole problem here is an edge case, and before I actually implemented any changes/fixes I wrote a integration tests that actually triggered the problem (made it reproducible), this fact already disproves you statement.

spape:
@fpellanda Can you also explain, how you would approach a Unit Test for a private method?!?

spape:
As discussed in person. Will write a unit tests, that also tries to explain those steps that are performed in order to get the final request options. 👍 @fpellanda

PR #210, opened 2017-03-28
Add missing seconds to regex for datetime detection

scan:
Java backends omit the seconds on ISO8601 times if they're `00`. The RegEx used to detect if it's a `Date` or `DateTime` did not take that into account and has been fixed.

### DIFF
--- a/spec/dummy/config/environments/test.rb
+++ b/spec/dummy/config/environments/test.rb
@@ -14,7 +14,7 @@ Rails.application.configure do
 
   # Configure static asset server for tests with Cache-Control for performance.
   config.serve_static_assets  = true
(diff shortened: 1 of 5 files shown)


spec/dummy/config/environments/test.rb
```diff
@@ -13,8 +13,8 @@
   config.eager_load = false
 
   # Configure static asset server for tests with Cache-Control for performance.
-  config.serve_static_assets  = true
-  config.static_cache_control = 'public, max-age=3600'
+  config.serve_static_assets        = true
```
Averethel: I think you also want to replace this line with:
```
config.public_file_server.enabled = true
```
if you're taking care of deprecation warnings

PR #214, opened 2017-05-05
Consider records with an empty href not persisted

crackofdusk:
*Patch version*

If the `href` attribute comes from a rails form, it will be `""` (an empty string) instead of nil. In that case, the record is still not persisted.

### DIFF
--- a/spec/record/persisted_spec.rb
+++ b/spec/record/persisted_spec.rb
@@ -12,7 +12,7 @@ describe LHS::Record do
     end
 
     context 'for new record' do
-      subject { Feedback.new }
+      subject { Feedback.new(href: '') }
 
       it 'is false' do
         expect(subject.persisted?).to be(false)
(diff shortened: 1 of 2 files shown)


spape:
Suggesting Patch Version

spec/record/persisted_spec.rb
```diff
@@ -12,7 +12,7 @@ class Feedback < LHS::Record
     end
 
     context 'for new record' do
-      subject { Feedback.new }
+      subject { Feedback.new(href: '') }
```
spape: Can we handle/cover both cases, `nil` and `href:''` ?

crackofdusk: Sounds good!

crackofdusk:
Thank you I meant a patch version.

Test added in https://github.com/local-ch/lhs/pull/214/commits/e06105d6105c99d3d4a33e2a63b9ce1bbd68f5b4

Issue #215, opened 2017-05-10
Nested collection thinks it's an item


Averethel:
I have the following case:

```ruby
rs = Restaurant.includes(place: :feedbacks).where(...)
fs = rs.first.place.feedbacks
# =>
# Data of Restaurant 
# > place > feedbacks
# :href => "xxx"
# :items => [{:href=> "xxx" } ... ]
fs.item?
# =>
true
```

check free-table-search app / feedback-migration branch

spape:
Why it happens:
While navigating/accessing data, LHS applies configuration from last recognized Record.

When drilling down from Restaurant, which is a search results and identifies collections with `docs` as configured in `Restaurant` it applies this configuration when accessing `feedbacks` to.

As feedbacks has `items` do identify collections, LHS does not recognize it as collection.

In a patch I will try to identify Records when accessing collections, what was intendend by you when adding `:datastore/feedbacks` endpoint to the `Feedback` record.

Right now, only items are identified/recognize when accessing.

spape:
Fixed in https://github.com/local-ch/lhs/releases/tag/v11.0.2

PR #216, opened 2017-05-15
Fix identifying records

spape:
*PATCH VERSION*

This change fixes three issues. All three in combination lead to the problem, that a nested data collection was not recognized as such, as mentioned in https://github.com/local-ch/lhs/issues/215

(1) When accessing data, items/collections only have been distinguished by being either a hash or an array. I fixed/improved that, so that also hashes with an `items_key` get recognized as a collection when accessed.

(2) When a nested resource is linked, and this link contained get parameters, those get parameters have been dropped as part of making the link dynamically overloadable or parameters of the url , according to defined endpoint templates, exchangeable. Those preset get-paramters are now kept/preserved.

(3) When accessing a collection, the returned data does not have been wrapped in an identified record. As LHS always applies configuration from parent records, accessing collections e.g. via configured parents, had LHS apply the parents configuration also on the nested data.

Example:
```
{
  docs: [
    place: { 
      href: 'http://local.ch/places/1',
      feedbacks: {
        href: 'http://local.ch/places/1/feedbacks',
        items: [{ review: 'great food' }]
      }
    }
  ]
}
```

As the parent record was configured to access collections via  `docs` it also has tried to identify the feedbacks collection by this key.

### DIFF
--- a/lib/lhs/concerns/proxy/accessors.rb
+++ b/lib/lhs/concerns/proxy/accessors.rb
@@ -24,11 +24,9 @@ class LHS::Proxy
       end
 
       record = LHS::Record.for_url(value[:href]) if value.is_a?(Hash) && value[:href]
-      if accessing_item?(value, record)
-        return access_item(value, record)
-      elsif accessing_collection?(value, record)
-        return access_collection(value, record)
-      end
+      
+      access_item(value, record) ||
+      access_collection(value, record) ||
       convert(value)
     end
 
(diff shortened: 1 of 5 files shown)


lib/lhs/concerns/proxy/accessors.rb
```diff
@@ -23,11 +23,28 @@ def get(name, *args)
         value = _data._raw[name.to_s.classify.to_sym] if value.nil?
       end
 
-      return access_item(value) if value.is_a?(Hash)
-      return access_collection(value) if value.is_a?(Array)
+      record = LHS::Record.for_url(value[:href]) if value.is_a?(Hash) && value[:href]
+      if accessing_item?(value, record)
```
fpellanda: Hm, I would recommend to keep the `accessing_item?` and `accessing_collection?` logic in the `access_item` and `access_collection` method and return nil if access_item or access_collection is not appropriate.

Then you can:
``` ruby
access_item(value, record) || access_collection(value,record) || convert(value)
```

And I guess you can write the methods a bit more readable.

Averethel: @fpellanda Both are okay, but I'd prefer single responsibility.
Also, IMO chained ifs are a bit more readable/understandable extendable than too many chained ors.

Issue #217, opened 2017-05-15
LHS iterate not properly when there are limit/offset parameters in href already


fpellanda:
It appends limit and offset parameter instead of overwrite
`...?offset=0&limit=10&limit=10&offset=10`
Log:
```
ETHON: performed EASY effective_url=http://datastore-stg/v2/places/iIJUNDkoWuI41kI1Kuw77A/available-assets?offset=0&limit=10&limit=100 response_code=200 return_code=ok total_time=0.353035
ETHON: started MULTI
ETHON:         performed EASY effective_url=http://datastore-stg/v2/places/iIJUNDkoWuI41kI1Kuw77A/available-assets?offset=0&limit=10&limit=10&offset=10 response_code=200 return_code=ok total_time=0.333243
```

PR #218, opened 2017-05-17
Fix `includes_all` for links with get params for pagination (like limit & offset)

spape:
Fixes https://github.com/local-ch/lhs/issues/217

### DIFF
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -292,8 +292,12 @@ class LHS::Record
       # as we have to fetch all resources on this level first, before we continue_including
       def prepare_option_for_include_all_request!(option)
         return option unless option.present?
+        uri = URI.parse(option[:url])
+        get_params = Rack::Utils.parse_nested_query(uri.query).symbolize_keys.except(limit_key, pagination_key)
         option[:params] ||= {}
+        option[:params].merge! get_params if get_params
         option[:params].merge!(limit_key => option.fetch(:params, {}).fetch(limit_key, LHS::Pagination::Base::DEFAULT_LIMIT))
+        option[:url] = option[:url].gsub("?#{uri.query}", '')
         option.delete(:including)
         option.delete(:referencing)
         option
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -234,7 +234,7 @@ class LHS::Record
         record = record_for_options(options) || self
         options = convert_options_to_endpoints(options) if record_for_options(options)
         begin
-          prepare_options_for_include_request!(options, sub_includes, references)
+          prepare_options_for!(:include_request, options, sub_includes, references)
           if references && references[:all] # include all linked resources
             load_include_all!(options, record, sub_includes, references)
           else # simply request first page/batch
@@ -246,7 +246,7 @@ class LHS::Record
       end
 
       def load_include_all!(options, record, sub_includes, references)
-        prepare_options_for_include_all_request!(options)
+        prepare_options_for!(:include_all_request, options)
         data = load_all_included!(record, options)
         references.delete(:all) # for this reference all remote objects have been fetched
         continue_including(data, sub_includes, references)
@@ -277,15 +277,15 @@ class LHS::Record
         !!(raw.is_a?(Hash) && raw[total_key] && raw[pagination_key])
       end
 
-      def prepare_options_for_include_all_request!(options)
+      def prepare_options_for!(prepare_method, options, *args)
         if options.is_a?(Array)
           options.each do |option|
-            prepare_option_for_include_all_request!(option)
+            send("prepare_option_for_#{prepare_method}!", option, *args)
           end
         else
-          prepare_option_for_include_all_request!(options)
+          send("prepare_option_for_#{prepare_method}!", options, *args)
         end
-        options
+        options || {}
       end
 
       # When including all resources on one level, don't forward :includes & :references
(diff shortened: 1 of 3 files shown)


lib/lhs/concerns/record/request.rb
```diff
@@ -234,7 +235,7 @@ def load_include(options, data, sub_includes, references)
         record = record_for_options(options) || self
         options = convert_options_to_endpoints(options) if record_for_options(options)
         begin
-          prepare_options_for_include_request!(options, sub_includes, references)
+          prepare_options_for!(:include_request, options, sub_includes, references)
```
spape: There is is only one `prepare_method` for now. Please stay with a concrete method, without dynamics, as we loose the benefit of "no method error" when called for unknown `perpare_method`. We can refactor that, as soon as we have multiple prepare scenarios... or we do it in a PR that just refactors `prepare_options_for`.

lib/lhs/concerns/record/request.rb
```diff
@@ -277,35 +278,38 @@ def paginated?(raw)
         !!(raw.is_a?(Hash) && raw[total_key] && raw[pagination_key])
       end
 
-      def prepare_options_for_include_all_request!(options)
+      def prepare_options_for!(prepare_method, options, *args)
         if options.is_a?(Array)
           options.each do |option|
-            prepare_option_for_include_all_request!(option)
+            send("prepare_option_for_#{prepare_method}!", option, *args)
```
spape: Im really not fan of that. Wouldn't find this call when searching for `prepare_option_for_ include_all_request!`

lib/lhs/concerns/record/request.rb
```diff
@@ -277,35 +278,38 @@ def paginated?(raw)
...
+          .except(limit_key, pagination_key)
         option[:params] ||= {}
-        option[:params].merge!(limit_key => option.fetch(:params, {}).fetch(limit_key, LHS::Pagination::Base::DEFAULT_LIMIT))
+        option[:params].reverse_merge!(get_params)
+        option[:params][limit_key] ||= LHS::Pagination::Base::DEFAULT_LIMIT
+        option[:url] = option[:url].gsub("?#{uri.query}", '')
         option.delete(:including)
         option.delete(:referencing)
         option
+      rescue URI::InvalidURIError
```
spape: Such errors have to be raised, so that the consumer understands what's going on. Imagine somebody trying to include something, and then has a hard time to understand, why it's just not including what was intended to be included, without further notice.

lib/lhs/concerns/record/request.rb
```diff
@@ -4,6 +4,7 @@ class LHS::Record
 
   module Request
     extend ActiveSupport::Concern
+    include Configuration
```
spape: Please don't include modules through modules!

spec/concerns/record/request_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'rails_helper'
+
+describe LHS::Record::Request do
+  subject { Class.new { include(LHS::Record::Request) } }
+
+  describe 'prepare_options_for!' do
+    it 'calls correct prepare method for nil' do
+      expect(subject).to receive(:prepare_option_for_testing!)
+        .with(nil, 'arg').and_return('ignore')
+      expect(subject.send(:prepare_options_for!, :testing, nil, 'arg')).to eq({})
```
spape: That's exactly not the behaviour I want it to raise, as `:testing` can't be prepared. I want to raise as not_method error. As I don't want to add/maintain custom error raising behaviour here.

spec/concerns/record/request_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  describe 'prepare_option_for_incldue_all_request' do
+    it 'removes all limit and offset parameters' do
+      option = { url: 'http://localhost:3000/test/resource?abc=def&limit=1&offset=3&test=2' }
+      expect(subject.send(:prepare_option_for_include_all_request!, option))
+        .to eq(option)
+      expect(option).to eq(url: 'http://localhost:3000/test/resource',
+                           params: { limit: 100, abc: 'def', test: '2' })
+    end
+
+    it 'do nothing without url' do
```
spape: Should raise.

lib/lhs/concerns/record/request.rb
fpellanda: Why not?
See: http://api.rubyonrails.org/classes/ActiveSupport/Concern.html
| Moreover, it gracefully handles module dependencies. 
It's a dependency of Request. Without this concern the request concern does not work.

lib/lhs/concerns/record/request.rb
fpellanda: Ok, I'll remove it again. I just don't like duplication...

spape:
I like how `request_spec.rb` tests the module more in detail and therefor makes the `prepare_options` methods easier understandable.

spec/concerns/record/request_spec.rb
fpellanda: Well, when there's a faulty url, raise. But this case should not. Otherwise we have to change this spec. https://github.com/local-ch/lhs/blob/master/spec/record/includes_all_spec.rb#L151
See: https://ci.local.ch/cider-ci/ui/workspace/trials/45e387c1-e512-4786-9e92-30d9193ed110/scripts/rspec
OK?

lib/lhs/concerns/record/request.rb
spape: It "gracefully" handles module dependencies means, it's a mess. Otherwise it would be "It handles module dependencies in a clear way.

spape: Because modules can overwrite functions depending on when the module was loaded by the ruby interpreter, that's super complicated and annoying to debug.

spec/concerns/record/request_spec.rb
spape: Ok. Only raise for faulty url, not for not present URL 👍

PR #219, opened 2017-05-30
Forward errors to nested data

spape:
**Minor Version**

### Validation errors for nested data

If you work with complex data structures, you sometimes need to have validation errors delegated/scoped to nested data.

This also makes LHS::Records compatible with how Rails or Simpleform renders/builds forms and especially error messages.

```ruby
# controller.rb
unless @customer.save
  @errors = @customer.errors
end

# view.html
= form_for @customer, as: :customer do |customer_form|
  
  = fields_for 'customer[:address]', @customer.address, do |address_form|
  
    = fields_for 'customer[:address][:street]', @customer.address.street, do |street_form|

      = street_form.input :name
      = street_form.input :house_number
```

Would render nested forms and would also render nested form errors for nested data structures.

You can also access those nested errors like:

```ruby
@customer.address.errors
@customer.address.street.errors
```


lib/lhs/concerns/proxy/errors.rb
```diff
@@ -0,0 +1,21 @@
...
+      attr_writer :errors
+    end
+
+    def initialize(data)
+      super(data)
+      self.errors = LHS::Errors::Base.new
+    end
+
+    def errors
+      @errors || @errors = LHS::Errors::Base.new
```
scan: What's the difference to using `||=` here?

lib/lhs/concerns/proxy/accessors.rb
```diff
@@ -54,20 +53,23 @@ def convert(value)
...
+      wrap_return(collection, record, name)
     end
 
-    def wrap_return(value, record)
-      data = LHS::Data.new(value, _data)
+    # Wraps with record and adds nested errors to data,
+    # if errors are existing
+    def wrap_return(value, record, name)
+      data = value.is_a?(LHS::Data) ? value : LHS::Data.new(value, _data)
+      data.errors = LHS::Errors::Nested.new(errors, name) if errors
```
ryabrody: Where is `errors` defined?

cedricwider:
Maybe consider fixing the specs...? ;)

cedricwider:
Maybe consider fixing the specs...? ;)

lib/lhs/concerns/proxy/accessors.rb
spape: https://github.com/local-ch/lhs/blob/forward-errors-to-nested-data/lib/lhs/concerns/proxy/errors.rb @ryabrody

lib/lhs/concerns/proxy/errors.rb
spape: 👍 refactored to use `||=`

PR #221, opened 2017-06-02
Validation errors translated by convention

spape:
*Minor Version*

### Translation of validation errors

Just like Activerecord, LHS tries to translate validation error messages.
If a translation exists for one of the following translation keys, LHS will take a translated error (also in the following order) rather than the plain error message/code:

```ruby
lhs.errors.records.customers.attributes.name.unsupported_property_value
lhs.errors.records.customers.unsupported_property_value
lhs.errors.messages.unsupported_property_value
lhs.errors.attributes.name.unsupported_property_value
lhs.errors.fallback_message
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -777,8 +777,8 @@ Just like Activerecord, LHS tries to translate validation error messages.
 If a translation exists for one of the following translation keys, LHS will take a translated error (also in the following order) rather than the plain error message/code:
 
 ```ruby
-lhs.errors.records.customers.attributes.name.unsupported_property_value
-lhs.errors.records.customers.unsupported_property_value
+lhs.errors.records.customer.attributes.name.unsupported_property_value
+lhs.errors.records.customer.unsupported_property_value
 lhs.errors.messages.unsupported_property_value
 lhs.errors.attributes.name.unsupported_property_value
 lhs.errors.fallback_message
--- a/README.md
+++ b/README.md
@@ -771,6 +771,19 @@ You can also access those nested errors like:
 @customer.address.street.errors
 ```
 
+### Translation of validation errors
+
+Just like Activerecord, LHS tries to translate validation error messages.
+If a translation exists for one of the following translation keys, LHS will take a translated error (also in the following order) rather than the plain error message/code:
+
+```ruby
+lhs.errors.records.customers.attributes.name.unsupported_property_value
+lhs.errors.records.customers.unsupported_property_value
+lhs.errors.messages.unsupported_property_value
+lhs.errors.attributes.name.unsupported_property_value
+lhs.errors.fallback_message
+```
+
 ### Know issue with `ActiveModel::Validations`
 If you are using `ActiveModel::Validations` and add errors to the LHS::Record instance - as described above - then those errors will be overwritten by the errors from `ActiveModel::Validations` when using `save`  or `valid?`. [Open issue](https://github.com/local-ch/lhs/issues/159)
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -771,6 +771,19 @@ You can also access those nested errors like:
 @customer.address.street.errors
 ```
 
+### Translation of validation errors
+
+Just like Activerecord, LHS tries to translate validation error messages.
+If a translation exists for one of the following translation keys, LHS will take a translated error (also in the following order) rather than the plain error message/code:
+
+```ruby
+lhs.errors.records.customers.attributes.name.unsupported_property_value
```
crackofdusk: The record name in these examples should be singular (customer) as per your implementation and the convention in Activerecord.

spape: Good catch 👍  Fixed it.

lib/lhs/errors/base.rb
```diff
@@ -76,7 +77,23 @@ def empty?
     def add_error(messages, key, value)
       key = key.to_sym
       messages[key] ||= []
-      messages[key].push(value)
+      messages[key].push(generate_message(key, value))
+    end
+
+    def generate_message(attribute, message, _options = {})
+      record_name = record.model_name.name.underscore
+      normalize_attribute = attribute.to_s.underscore
```
cedricwider: should probably be `normalized_attribute` (with a d)

lib/lhs/errors/base.rb
```diff
@@ -76,7 +77,23 @@ def empty?
...
+      normalized_message = message.to_s.underscore
+      [
+        ['lhs', 'errors', 'records', record_name, 'attributes', normalize_attribute, normalized_message],
+        ['lhs', 'errors', 'records', record_name, normalized_message],
+        ['lhs', 'errors', 'messages', normalized_message],
+        ['lhs', 'errors', 'attributes', normalize_attribute, normalized_message],
+        ['lhs', 'errors', 'fallback_message']
+      ].detect do |path|
```
cedricwider: This seems to be a concept. Maybe extract method and add a name (like 'find error key') to this concept?

README.md
```diff
@@ -771,6 +771,19 @@ You can also access those nested errors like:
...
+
+Just like Activerecord, LHS tries to translate validation error messages.
+If a translation exists for one of the following translation keys, LHS will take a translated error (also in the following order) rather than the plain error message/code:
+
+```ruby
+lhs.errors.records.customer.attributes.name.unsupported_property_value
+lhs.errors.records.customer.unsupported_property_value
+lhs.errors.messages.unsupported_property_value
+lhs.errors.attributes.name.unsupported_property_value
+lhs.errors.fallback_message
```
ryabrody: Why do we not support the error messages from the rails.yml as example: `errors.messages.invalid` in here?

lib/lhs/errors/base.rb
spape: true

lib/lhs/errors/base.rb
spape: refactored 👍

README.md
spape: Because APIs at local will not send `invalid` or follow Rails conventions, like all the messages in that file. In this regard, we have to adapt to API keys, rather then API adapt to some Rails conventions.

spape: Or do you mean, why we use `lhs.errors.fallback_message` instead of `invalid` ?

spape: Please add an follow up issue and we will change, add it incrementally.

PR #222, opened 2017-06-07
Fix: There is no record for nested unknown data types

spape:
*PATCH VERSION*

## BEFORE
```
undefined method `model_name' for nil:NilClass
lhs (11.2.0) lib/lhs/errors/base.rb:88:in `find_translated_error_message'
lhs (11.2.0) lib/lhs/errors/base.rb:84:in `generate_message'
lhs (11.2.0) lib/lhs/errors/base.rb:35:in `set'
lhs (11.2.0) lib/lhs/errors/base.rb:41:in `[]'
```

## NOW
Smooth as silk

### DIFF
--- a/lib/lhs/errors/base.rb
+++ b/lib/lhs/errors/base.rb
@@ -85,19 +85,26 @@ module LHS::Errors
     end
 
     def find_translated_error_message(attribute, message)
-      record_name = record ? record.model_name.name.underscore : nil
       normalized_attribute = attribute.to_s.underscore
       normalized_message = message.to_s.underscore
-      [
-        record_name ? ['lhs', 'errors', 'records', record_name, 'attributes', normalized_attribute, normalized_message] : nil,
-        record_name ? ['lhs', 'errors', 'records', record_name, normalized_message] : nil,
+      messages = []
+      messages = messages_for_record(normalized_attribute, normalized_message) if record
+      messages.concat([
         ['lhs', 'errors', 'messages', normalized_message],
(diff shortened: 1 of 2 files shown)


lib/lhs/errors/base.rb
```diff
@@ -85,16 +85,16 @@ def generate_message(attribute, message, _options = {})
...
     def find_translated_error_message(attribute, message)
-      record_name = record.model_name.name.underscore
+      record_name = record ? record.model_name.name.underscore : nil
       normalized_attribute = attribute.to_s.underscore
       normalized_message = message.to_s.underscore
       [
-        ['lhs', 'errors', 'records', record_name, 'attributes', normalized_attribute, normalized_message],
-        ['lhs', 'errors', 'records', record_name, normalized_message],
+        record_name ? ['lhs', 'errors', 'records', record_name, 'attributes', normalized_attribute, normalized_message] : nil,
+        record_name ? ['lhs', 'errors', 'records', record_name, normalized_message] : nil,
```
fpellanda: Don't like the way you're using the `?`.  I'd like more when you put an if around the code which is record related. (Instead of these `?` and `nil` and `compact`...

```
if record
  record_name = record.model_name.name.underscore
  ...['lhs', 'errors', 'records', record_name, 'attributes', normalized_attribute, normalized_message]
  ...['lhs', 'errors', 'records', record_name, normalized_message]
end
```

spape: Refactored

PR #223, opened 2017-06-15
Fix url parsing for records using host placeholders

spape:
*PATCH VERSION*

Problem:

```ruby
class Entry < LHS::Record
  endpoint ':datastore/entry/v1/:id.json'
end
``` 

Records configured with host placeholders (see above) where raising URL.parsing exceptions, as they are per se not real urls by any standards.

Solution:

Computing urls from parameters when placeholders are detected, before the urls are actually parsed.

### DIFF
--- a/spec/record/includes_all_spec.rb
+++ b/spec/record/includes_all_spec.rb
@@ -212,7 +212,7 @@ describe LHS::Record do
       end
     end
 
-    context 'include a known/identifyable record' do
+    context 'include a known/identifiable record' do
       before(:each) do
         class Contract < LHS::Record
           endpoint 'http://datastore/contracts/:id'
@@ -253,7 +253,7 @@ describe LHS::Record do
           )
       end
 
-      it 'loads included identifyable records withou raising exceptions' do
+      it 'loads included identifiable records without raising exceptions' do
         customer = Customer.includes_all(contracts: :entry).find(1, 2).first
         expect(customer.contracts.first.href).to eq 'http://datastore/contracts/1'
         expect(customer.contracts.first.type).to eq 'contract'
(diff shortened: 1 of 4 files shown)


spec/record/includes_all_spec.rb
```diff
@@ -211,5 +211,54 @@ class Contract < LHS::Record
...
+      let!(:entry_request) do
+        stub_request(:get, %r{http://datastore/entry/v1/\d+.json})
+          .to_return(
+            body: {
+              name: 'Casa Ferlin'
+            }.to_json
+          )
+      end
+
+      it 'loads included identifyable records withou raising exceptions' do
```
crackofdusk: Typos:
identif_i_able
 withou_t_

spec/record/includes_all_spec.rb
```diff
@@ -211,5 +211,54 @@ class Contract < LHS::Record
         end
       end
     end
+
+    context 'include a known/identifyable record' do
```
crackofdusk: Typo: identif_i_able

spec/record/includes_all_spec.rb
```diff
@@ -211,5 +211,54 @@ class Contract < LHS::Record
...
+            }.to_json
+          )
+      end
+
+      let!(:contracts_request) do
+        stub_request(:get, %r{http://datastore/contracts/\d+})
+          .to_return(
+            body: {
+              type: 'contract',
+              entry: { href: 'http://datastore/entry/v1/1.json' }
```
ryabrody: Just a general question: When does a user need to add `json` prefix?

spape: When the endpoint does not support correct acceptance headers for example and only relies on format in the url (which is not recommended anyway)

spec/record/includes_all_spec.rb
spape: incred_i_ble 
thanks ;)

spape:
Fixed the issues in SB too ;)

PR #224, opened 2017-06-19
Fix weird order for special `includes_all` case

spape:
*PATCH*

### PROBLEM

Imagine you work with places, and a place has `category_relations`
```
{
  id: 1,
  category_relations: [ { href: 'http://datastore/category_relations/1' } ]
}
```

When multiple places have been requested, with including all their `category_relations`, places with an empty array of `category_relations` where messing up the order of extended data, which lead to weird merged state of the places objects in the end...

```
{
  id: 1,
  category_relations: [ { href: 'http://datastore/category_relations/1' } ]
}

{
  id: 2,
  category_relations: []
}

{
  id: 3,
  category_relations: [ { href: 'http://datastore/category_relations/2' }, { href: 'http://datastore/category_relations/3' } ]
}
```

Was creating a request object array looking like:
```
[ 
  { url: 'http://datastore/category_relations/1' },
  nil,
  { url: 'http://datastore/category_relations/2' },
  { url: 'http://datastore/category_relations/3' }
]
```

Because the merge is depending on the additions index, the nil was messing up and shifting everything afterwards by 1.

### FIX

Filter all those nils in the request object array, in order to prevent shifts of merged data.

### DIFF
--- a/lib/lhs/collection.rb
+++ b/lib/lhs/collection.rb
@@ -33,13 +33,13 @@ class LHS::Collection < LHS::Proxy
     true
   end
 
-  def compact
-    LHS::Collection.new(
-      LHS::Data.new(
-        _collection.raw.reject { |item| item.nil? }
-      )
-    )
-  end
+  # def compact
+  #   LHS::Collection.new(
+  #     LHS::Data.new(
+  #       _collection.raw.reject { |item| item.nil? }
+  #     )
+  #   )
+  # end
 
   def item?
     false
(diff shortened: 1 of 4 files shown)


fpellanda:
@spape Let's sit together again. Your solution did not help. However I found a solution but have no idea if this is correct. But I know "where bunny lies in the pepper".

spape:
@fpellanda This PR was not reviewable, yet. Hat a solution locally already. Pushed it now. Reverted your's while merging.

spape:
If you want to come up with an own solution, please make a separated PR @fpellanda

PR #228, opened 2017-06-22
WIP: Add spec to reproduce error from UFO-493

spape:


cedricwider:
Deleting branch as it only ever served demo purposes

Issue #213, opened 2017-04-24
Load gem files: Dir[].each vs. Gem.find_files


spape:
Currently we partially have deprecation warnings, because in some cases we load multiple lhc versions (like in development).

http://stackoverflow.com/questions/735073/best-way-to-require-all-files-from-a-directory-in-ruby

PR #230, opened 2017-06-28
Autoload for the win

spape:
*MAJOR*

Implements the proper way of loading dependencies in ruby, the `autoload` way (as seen in famous gems like [Bundler](https://github.com/bundler/bundler/blob/master/lib/bundler.rb))

Also Fixes: https://github.com/local-ch/lhs/issues/213


lib/lhs/concerns/autoload_records.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+
+        def call(env)
+          self.class.require_records
+          @app.call(env)
+        end
+
+        def self.require_records
+          Dir.glob(Rails.root.join('app/models/**/*.rb')).each do |file|
+            require_dependency file if File.read(file).match('LHS::Record')
```
Averethel: Shouldn't we match for inheritance from LHS::Record?

spape: true, but I just moved code. Please make an Issue or a PR.

PR #231, opened 2017-06-29
Fix error access with indifferent access

spape:
*PATCH*

Access error messages hash was not able with indifferent access.

I found out that there such an access in location app: e.g. https://github.com/local-ch/location-app/blob/test-lhs-autoload/app/views/restaurants_feedbacks/_review_form.html.haml#L13

This change makes accessing error messages possible also for indifferent access.

### DIFF
--- a/spec/record/creation_failed_spec.rb
+++ b/spec/record/creation_failed_spec.rb
@@ -40,7 +40,7 @@ describe LHS::Record do
       expect(record.errors.include?(:ratings)).to eq true
       expect(record.errors.include?(:recommended)).to eq true
       expect(record.errors[:ratings]).to eq ['REQUIRED_PROPERTY_VALUE']
-      expect(record.errors.messages).to eq(ratings: ["REQUIRED_PROPERTY_VALUE"], recommended: ["REQUIRED_PROPERTY_VALUE"])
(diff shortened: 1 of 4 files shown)


spec/record/creation_failed_spec.rb
```diff
@@ -40,7 +40,7 @@ class Record < LHS::Record
       expect(record.errors.include?(:ratings)).to eq true
       expect(record.errors.include?(:recommended)).to eq true
       expect(record.errors[:ratings]).to eq ['REQUIRED_PROPERTY_VALUE']
-      expect(record.errors.messages).to eq(ratings: ["REQUIRED_PROPERTY_VALUE"], recommended: ["REQUIRED_PROPERTY_VALUE"])
+      expect(record.errors.messages).to eq('ratings' => ["REQUIRED_PROPERTY_VALUE"], 'recommended' => ["REQUIRED_PROPERTY_VALUE"])
```
Averethel: `'ratings': ["REQUIRED_PROPERTY_VALUE"], 'recommended': ["REQUIRED_PROPERTY_VALUE"]` should work too no?

spape: nope 
```ruby
2.3.0 :002 > {'key': 2}
 => {:key=>2}
```

```ruby
2.3.0 :003 > {'key' => 2}
 => {"key"=>2} 
```

PR #235, opened 2017-07-26
Make nested items and pagination meta data accessible with complex accessors.

spape:
_MINOR_

LHS is stable, but this feature in particular is to be considered "experimental".

### Configure complex accessors for nested data (EXPERIMENTAL)

If items, limit, pagination, total etc. is nested in the responding objects, use the complex data structures to for configuring your record.

```
  response: {
    offset: 0,
    max: 50,
    count: 1,
    businesses: [
      {}
    ]
  }
```

```ruby
  class Business < LHS::Record
    configuration items_key: [:response, :businesses], limit_key: [:response, :max], pagination_key: [:response, :offset], total_key: [:response, :count], pagination_strategy: :offset
    endpoint 'http://uberall/businesses'
  end
```


README.md
```diff
@@ -865,7 +876,27 @@ end
 
 `total_key` key used to determine the total amount of items (e.g. `total`, `totalResults`, etc.).
 
-In case of paginated resources it's important to know the difference between [count vs. length](#count-vs-length)
+### Configure complex accessors for nested data (EXPERIMENTAL)
+
+If items, limit, pagination, total etc. is nested in the responding objects, use the complex data structures to for configuring your record.
```
ryabrody: rm "to"
"... use the complex data structures ---> to <--- for configuring..."

spape: 👍

fpellanda:
Just wondering why adding an experimental feature to upstream? Who will use this? Maybe mark somehow as experimental?

spape:
@fpellanda 
Uberall's API has complex accessors. As we cant make them change their APIs as we would at local, I have to start implementing/using complex accessors. It's common practice to just mention functionalities in an experimental state in the documentation, as this would be the first place, people consuming new interfaces of LHS are looking at anyway. We could mentioned it in the source code code, but who is checkin source code anyway? Spitting out warnings might be a little much.

This new functionality is not worse in any sense as previous functionalities. Just from experience I can tell, that such new functionalities have some "teething problems", I plan to remove the "EXPIREMENTAL" tag from the documentation as soon as I feel comfortable with the functionality, but that's totally a gut feeling.

lib/lhs/concerns/items_handler.rb
```diff
@@ -0,0 +1,29 @@
+require 'active_support'
+
+module LHS
+  module ItemsHandler
```
crackofdusk: All the methods in this module reference `items`. There may be a missing concept here.

spape: Improved the concept.

The logic about handling potentially (deep-)nested collections lives now in a collection concern called `HandleNested` (see: https://github.com/local-ch/lhs/pull/235/commits/143ac9dba0ff5b2e0e79beb14f2c45d2ab45a812#diff-a963c753f4ea79c01424fe1004fa5066).

I also tried to find better names for the methods and added some documentation to the concern and methods.

I also added a instance > class pipe-trough, to keep the convince I told you about, to still be able to access stuff from the instance perspective.

Would be glad if you could check the changes.

👍

PR #240, opened 2017-08-11
bump required ruby to 2.3.0, because of use of #dig

skaestle:


spape:
Major, I would suggest.

PR #241, opened 2017-08-11
Proxy#reload! also with id

spape:
_PATCH_

Due to the fact that within localsearch we heavily rely on hrefs, `reload!` was not take in consideration to reload records by present id, wich can be necessary for other APIs, like uberall.

### DIFF
--- a/lib/lhs/proxy.rb
+++ b/lib/lhs/proxy.rb
@@ -17,6 +17,7 @@ class LHS::Proxy
 
   # prevent clashing with attributes of underlying data
   attr_accessor :_data, :_loaded
+  delegate :_record, to: :_data, allow_nil: true
 
   def initialize(data)
     self._data = data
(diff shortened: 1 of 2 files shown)


lib/lhs/proxy.rb
```diff
@@ -33,12 +33,20 @@ def load!(options = nil)
...
     _data.merge_raw!(data)
     self._loaded = true
     self
   end
+
+  private
+
+  def reload_options
+    return { url: _data.href } if _data.href
+    return { params: { id: _data.id } } if _data.id
```
fpellanda: Can't we just overwrite this somehow in the used class? Here we again implement something API specific for uberall.

```ruby
class Location
  def reload_options
    { params: { id: self.id } }
  end
end
```

spape: Well. From the code change, you cant really tell it's uberall specific. But yeah, this idea sounds reasonable.

PR #244, opened 2017-08-16
.ignore(error)

spape:
If an error handler returns `nil` an empty LHS::Record is returned, not `nil`!

In case you want to ignore errores and continue working with `nil` in those cases,
please use `ignore`:

```ruby
record = Record.ignore(LHC::NotFound).find_by(color: 'blue')
record # nil
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ gem 'lhs'
 LHS comes with Request Cycle Cache – enabled by default. It requires [LHC Caching Interceptor](https://github.com/local-ch/lhc/blob/master/docs/interceptors/caching.md) to be enabled:
 
 ```ruby
-# intializers/lhc.rb 
+# intializers/lhc.rb
 LHC.configure do |config|
   config.interceptors = [LHC::Caching]
 end
@@ -295,7 +295,7 @@ You can apply options to the request chain. Those options will be forwarded to t
 
 ## Request Cycle Cache
 
-By default, LHS does not perform the same http request during one request cycle multiple times. 
+By default, LHS does not perform the same http request during one request cycle multiple times.
 
 It uses the [LHC Caching Interceptor](https://github.com/local-ch/lhc/blob/master/docs/interceptors/caching.md) as caching mechanism base and sets a unique request id for every request cycle with Railties to ensure data is just cached within one request cycle and not shared with other requests.
 
@@ -307,6 +307,12 @@ If you want to disable the LHS Request Cycle Cache, simply disable it within con
 
 ```ruby
 LHS.config.request_cycle_cache_enabled = false
+```
+
+By default the LHS Request Cycle Cache will use `ActiveSupport::Cache::MemoryStore` as its cache store. Feel free to configure a cache that is better suited for your needs by:
+
+```ruby
+LHS.config.request_cycle_cache = ActiveSupport::Cache::MemoryStore.new
 ```
 
 ## Batch processing
@@ -693,7 +699,7 @@ record.update(recommended: false)
 Based on [ActiveRecord's implementation](http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-becomes), LHS implements `becomes`, too.
 It's a way to convert records of a certain type A to another certain type B.
 
-_NOTE: RPC-style actions, that are discouraged in REST anyway, are utilizable with this functionality, too. See the following example:_ 
+_NOTE: RPC-style actions, that are discouraged in REST anyway, are utilizable with this functionality, too. See the following example:_
 
 ```ruby
 class Location < LHS::Record
@@ -802,9 +808,9 @@ end
 
 # view.html
 = form_for @customer, as: :customer do |customer_form|
-  
+
   = fields_for 'customer[:address]', @customer.address, do |address_form|
-  
+
     = fields_for 'customer[:address][:street]', @customer.address.street, do |street_form|
 
       = street_form.input :name
--- a/README.md
+++ b/README.md
@@ -181,6 +181,14 @@ record = Record.where(color: 'blue')
 
 [List of possible error classes](https://github.com/local-ch/lhc/tree/master/lib/lhc/errors)
 
+## Resolve chains
+
+LHS Chains can be resolved with `fetch`, similiar to ActiveRecord:
+
+```ruby
+records = Record.where(color: 'blue').fetch
+```
+
 ## Find single records
 
 `find` finds a unique record by unique identifier (usualy id or href).
--- a/README.md
+++ b/README.md
@@ -181,6 +181,16 @@ record = Record.where(color: 'blue')
 
 [List of possible error classes](https://github.com/local-ch/lhc/tree/master/lib/lhc/errors)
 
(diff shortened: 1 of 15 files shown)


lib/lhs/concerns/record/request.rb
```diff
@@ -9,21 +9,21 @@ module Request
...
       end
 
       private
 
-      def filter_request_options!(options)
-        options.each_with_index do |option, index|
-          next if !option || !option.key?(:url) || !option[:url].nil?
-          options[index] = nil
+      def filter_empty_request_options(options)
+        options.each_with_index.map do |option|
```
cedricwider: I assume you could just use `map` since you don't care about the index?

README.md
```diff
@@ -181,6 +181,16 @@ record = Record.where(color: 'blue')
 
 [List of possible error classes](https://github.com/local-ch/lhc/tree/master/lib/lhc/errors)
 
+If an error handler returns `nil` an empty LHS::Record is returned, not `nil`!
+
+In case you want to ignore errores and continue working with `nil` in those cases,
```
Averethel: `In case you want to ignore errors`

lib/lhs/concerns/record/request.rb
spape: We do care about the index. we need nils in the request and in the response array.

See: https://github.com/local-ch/lhs/blob/ignore-errors/lib/lhs/concerns/record/request.rb#L365

lib/lhs/concerns/record/chainable.rb
```diff
@@ -45,6 +45,10 @@ def handle(error_class, handler)
         Chain.new(self, ErrorHandling.new(error_class => handler))
       end
 
+      def ignore(error_class)
```
Averethel: Can we pass multiple errors here (`*error_classes`)?
It might be the case that you want to ignore more than just one error and it feels weird to write:
`Record.ignore(LHC::NotFound).ignore(LHC::Unauthorized)` instead of:
`Record.ignore(LHC::NotFound, LHC::Unauthorized)`

lib/lhs/concerns/record/request.rb
cedricwider: Ah, I see. Thanks for clarification

lib/lhs/concerns/record/chainable.rb
```diff
@@ -350,31 +363,37 @@ def push(link)
...
       end
 
       def chain_error_handler
-        _links.select { |link| link.is_a? ErrorHandling }
+        @chain_error_handler ||= _links.select { |link| link.is_a? ErrorHandling }
+      end
+
+      def chain_ignored_errors
+        @chain_ignored_errors ||= _links
+          .select { |link| link.is_a? IgnoredError }
```
Averethel: don't we also need to check the presence of `link.data`?

lib/lhs/concerns/record/chainable.rb
```diff
@@ -350,31 +363,37 @@ def push(link)
...
+          .map { |link| link.data }
       end
 
       def chain_pagination
-        resolve_pagination _links.select { |link| link.is_a? Pagination }
+        @chain_pagination ||= resolve_pagination _links.select { |link| link.is_a? Pagination }
```
Averethel: I would really appreciate parens here:
`resolve_pagination(_links.select { |link| link.is_a? Pagination })`
feels more readable than the original

lib/lhs/concerns/record/request.rb
Averethel: `map` seems to be enough

spec/record/ignore_errors_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    before(:each) do
+      stub_request(:get, "http://local.ch/v2/records?color=blue").to_return(status: 404)
+    end
+
+    it 'allows to ignore errors' do
+      record = Record
+        .where(color: 'blue')
+        .ignore(LHC::NotFound)
+        .fetch
+      expect(record).to eq nil
```
Averethel: `eq(nil)` for readability

spec/record/ignore_errors_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  context 'multiple ignored errors' do
+    it 'ignores error if one of them is specified' do
+      stub_request(:get, "http://local.ch/v2/records?color=blue").to_return(status: 401)
+      record = Record
+        .ignore(LHC::Unauthorized)
+        .where(color: 'blue')
+        .ignore(LHC::NotFound)
+        .fetch
+      expect(record).to eq nil
```
Averethel: `eq(nil)` for readability

spec/record/ignore_errors_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    it 'ignores error if one of them is specified' do
+      stub_request(:get, "http://local.ch/v2/records?color=blue").to_return(status: 404)
+      record = Record
+        .ignore(LHC::Unauthorized)
+        .where(color: 'blue')
+        .ignore(LHC::NotFound)
+        .fetch
+      expect(record).to eq nil
```
Averethel: `eq(nil)` for readability

spec/record/ignore_errors_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+  end
+
+  it 'also can ignore all LHC errors' do
+    stub_request(:get, "http://local.ch/v2/records?color=blue").to_return(status: 401)
+    record = Record
+      .ignore(LHC::Error)
+      .where(color: 'blue')
+      .fetch
+    expect(record).to eq nil
```
Averethel: `eq(nil)` for readability

lib/lhs/concerns/record/chainable.rb
```diff
@@ -193,6 +201,10 @@ def options(hash = nil)
         push(Option.new(hash))
       end
 
+      def ignore(error_class)
```
Averethel: Can we pass multiple errors here (*error_classes)?
It might be the case that you want to ignore more than just one error and it feels weird to write:
Record.ignore(LHC::NotFound).ignore(LHC::Unauthorized) instead of:
Record.ignore(LHC::NotFound, LHC::Unauthorized)

lib/lhs/concerns/record/request.rb
spape: Ah damn. Sorry @cedricwider understood your statement wrong. I thought we don't care about the `nils`. yeah, will drop the `with_index` @Averethel & @cedricwider 👍

README.md
spape: 👍

lib/lhs/concerns/record/chainable.rb
spape: 👍

lib/lhs/concerns/record/chainable.rb
spape: no. internal data structure. the would be ignored or raise on lhc level if data structure is incorrect.

lib/lhs/concerns/record/chainable.rb
spape: 👍

PR #245, opened 2017-08-17
use a configurable cache store for the request cycle cache

skaestle:
_PATCH_

fixes: https://jira.localsearch.ch/browse/WEB-4155 by introducing a configurable cache for the request cycle cache interceptor. 

it will try to use a `ActiveSupport::Cache::MemoryStore` if defined

### DIFF
--- a/spec/request_cycle_cache/main_spec.rb
+++ b/spec/request_cycle_cache/main_spec.rb
@@ -87,4 +87,17 @@ describe 'Request Cycle Cache', type: :request do
       expect(request).to have_been_made.times(2)
     end
   end
+
+  context 'use: cache' do
+    let!(:old_cache) { LHS.config.request_cycle_cache}
+    before { LHS.config.request_cycle_cache = double('cache') }
+    after { LHS.config.request_cycle_cache = old_cache }
+
+    it 'uses the cache passed in',
+    cleanup_before: true, request_cycle_cache: true do
+      expect(LHS.config.request_cycle_cache).to receive(:fetch).twice.and_return(nil)
+      expect(LHS.config.request_cycle_cache).to receive(:write).twice.and_return(nil)
+      get '/request_cycle_cache/simple'
+    end
+  end
 end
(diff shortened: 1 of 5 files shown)


spec/request_cycle_cache/main_spec.rb
```diff
@@ -87,4 +87,17 @@ class User < LHS::Record
...
   end
+
+  context 'use: cache' do
+    let!(:old_cache) { LHS.config.request_cycle_cache }
+    before { LHS.config.request_cycle_cache = double('cache') }
+    after { LHS.config.request_cycle_cache = old_cache }
+
+    it 'uses the cache passed in',
+    cleanup_before: true, request_cycle_cache: true do
+      expect(LHS.config.request_cycle_cache).to receive(:fetch).twice.and_return(nil)
```
crackofdusk: Since `and_return(nil)` is a technical detail rather than part of the verification step, maybe you can move the responses to the double?

```ruby
double('cache', fetch: nil, write: nil)
```

crackofdusk: How come we make one request but search the cache twice?

spec/request_cycle_cache/main_spec.rb
```diff
@@ -87,4 +87,17 @@ class User < LHS::Record
...
+      expect(LHS.config.request_cycle_cache).to receive(:write).twice.and_return(nil)
```
crackofdusk: Same as above: why do interact with the cache twice? Also, if an item is already in the cache, why do we overwrite it?

spec/request_cycle_cache/main_spec.rb
skaestle: ok so as far as I understand:

```ruby
def simple
  User.find(1) # first request
  user = User.find(1) # second request that should be serverd from request cycle cache
  render json: user.to_json
end
```

so on the first request we have a `fetch` (which detects no entry in the cache) and then `write`. on the second request we have another `fetch` (which also detects no entry - because the cache does actually not work as it's just a test double) so it continues on and writes it into the cache twice

skaestle: it might be a bit confusing, but the test does not actually test the caching feature itself but only if the cache that was passed in via the configuration was actually used.

skaestle:
@spape thanks! done

PR #248, opened 2017-08-18
add test-support for RequestCycleCache

skaestle:
_PATCH_

I noticed some weird test-fails on salesbutler (https://ci.local.ch/cider-ci/ui/workspace/jobs/6b3ac0dd-5c2e-481b-8761-000b512bf07a) after switching to the new implementation of RequestCycleCache. So I guess this is necessary?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1102,6 +1102,16 @@ it 'displays contracts' do
 end
 ```
 
+## Test support (caching)
+
+Add to your spec_helper.rb:
+
+```ruby
+  require 'lhs/test/request_cycle_cache.rb'
+```
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1102,6 +1102,16 @@ it 'displays contracts' do
 end
 ```
 
+## Test support (caching)
+
+Add to your spec_helper.rb:
+
+```ruby
+  require 'lhs/test/request_cycle_cache.rb'
```
crackofdusk: Isn't the idiom to omit the `.rb` suffix?

crackofdusk:
My bad that one is LHC and not in LHS. Ignore my comment above.

PR #250, opened 2017-08-21
Ignore errors improvements

spape:
Implements some feedback from @Averethel (crazy stuff)... https://github.com/local-ch/lhs/pull/244

Most important improvements is the ability to provide a list of ignored errors:

```ruby
Record
  .ignore(LHC::Unauthorized, LHC::Forbidden)
  .where(color: :blue)
```


Averethel:
If only you tagged right person @spape ;)

spape:
First time I type your name correctly without autocomplete or looking it up ;) @Averethel

Issue #249, opened 2017-08-21
Reload does not keep original record class


spape:
```ruby
record.reload!
```

Gives back an LHS::Item and not the Record instance any more.

See: https://github.com/local-ch/opm/pull/7/files#r134158846

spape:
Fixed in: https://github.com/local-ch/lhs/pull/251

PR #251, opened 2017-08-22
Reload returns instance of correct type now

spape:
_PATCH_

LHS was returning objects from type LHS::Item rather then the correct Record class on `reload!`.

Fixes: https://github.com/local-ch/lhs/issues/249

### Before

```ruby
record.reload! # LHS::Item
```

### Now

```ruby
record.reload! # LHS::Record
```

### DIFF
--- a/lib/lhs/proxy.rb
+++ b/lib/lhs/proxy.rb
@@ -40,6 +40,7 @@ class LHS::Proxy
     )
     _data.merge_raw!(data)
     self._loaded = true
+    return becomes(_record) if _record
     self
   end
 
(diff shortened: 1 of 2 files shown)


lib/lhs/proxy.rb
```diff
@@ -40,6 +40,7 @@ def reload!(options = nil)
     )
     _data.merge_raw!(data)
     self._loaded = true
+    return becomes(_record) if _record
```
ryabrody: why does record has a underline in its variable name?

spape: In libraries, that's indicating internal variables.

References:
https://stackoverflow.com/a/6700849
https://stackoverflow.com/a/150231

PR #259, opened 2017-09-18
Unwrap single items within nested response body

spape:
_MINOR_

## Configuration of Records

`item_key` key used to unwrap the actuall object from within the response body.
[...]

### Unwrap nested items

```json
{
  "response": {
    "location": {
      "id": 123
    }
  }
}
```

```ruby
class Location < LHS::Record
  configuration item_key: [:response, :location]
end

location = Location.find(1)
location.id # 123
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -914,6 +914,8 @@ class Search < LHS::Record
 end
 ```
 
+`item_key` key used to unwrap the actuall object from within the response body.
+
 `items_key` key used to determine items of the current page (e.g. `docs`, `items`, etc.).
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -926,6 +928,27 @@ end
...
+  }
+}
+```
+
+```ruby
+class Location < LHS::Record
+  configuration item_key: [:response, :location]
+end
+
+location = Location.find(1)
```
Averethel: Shouldn't this be 123?

PR #268, opened 2017-10-18
Separate pagination parameters

spape:
_MINOR_

Till now `pagination_key` and `limit_key` has been used for 2 usecases:
- find the current page and the current limit in the body
- provide a value for page and limit as get parameter when fetching multiple pages

Endpoints/Services like Uberall have 2 different keys/paths for those 2 use cases.

So in order to be able to use LHS for Uberall, I added the functionality to devide between those 2 use cases.

## Configuration of Records
[...]
`limit_key` key used to work with page limits (e.g. `size`, `limit`, etc.)

In case the `limit_key` parameter differs for where it's located in the body and how it's provided as get parameter, when retreiving pages, provide a hash with `body` and `paramter` key, to keep those two use cases separated:

```ruby
  configuration limit_key: { body: [:response, :max], parameter: :max }
```

`pagination_key` key used to paginate multiple pages (e.g. `offset`, `page`, `startAt` etc.).

In case the `pagination_key` parameter differs for where it's located in the body and how it's provided as get parameter, when retreiving pages, provide a hash with `body` and `paramter` key, to keep those two use cases separated:

```ruby
  configuration pagination_key: { body: [:response, :page], parameter: :page }
```
[...]

### DIFF
--- a/Gemfile.activesupport4
+++ b/Gemfile.activesupport4
@@ -1,4 +1,4 @@
 source 'https://rubygems.org/'
 
 gemspec
-gem 'activesupport', '~> 4.2.0'
+gem 'activesupport', '~> 4.2'
--- a/Gemfile.activesupport4
+++ b/Gemfile.activesupport4
@@ -1,4 +1,4 @@
 source 'https://rubygems.org/'
 
 gemspec
-gem 'activesupport', '~> 4.2'
+gem 'activesupport', '~> 4.2.0'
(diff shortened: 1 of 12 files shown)


Averethel:
Don't have time to fully review code right now, a question about backwards compatibility pops up in my head instantly when I read the description though...

spape:
@Averethel It's backwards compatible... didn't change any of the old tests.

PR #272, opened 2017-11-06
Fix require lhs

christianarevalo:
#PATCH

Issue: using `require 'lhs'` in a non rails app cause some issues with `delegate`.
`NoMethodError: undefined method `delegate' for LHS::Pagination::Base:Class`

This solves this specific use case.

### DIFF
--- a/spec/require_lhs_spec.rb
+++ b/spec/require_lhs_spec.rb
@@ -1,9 +1,9 @@
 require 'pry'
 require 'webmock/rspec'
 
-describe 'Require LHS' do
-  context 'request without rails' do
-    it 'does have deep_merge dependency met' do
+describe 'LHS' do
+  context 'when requiring lhs' do
+    it 'does not raise an exception' do
       expect { require 'lhs' }.not_to raise_error
     end
   end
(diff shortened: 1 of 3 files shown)


spec/require_lhs_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'pry'
+require 'webmock/rspec'
+
+describe 'Require LHS' do
+  context 'request without rails' do
+    it 'does have deep_merge dependency met' do
```
spape: I think this specification is copy pasted?

spec/require_lhs_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'pry'
```
skaestle: leftover?

christianarevalo: Actually I tried to require spec_helper as Sebastian suggested, but it was requiring some support files as well and thus breaking before the actual test. So for consistency I only require 'pry' and 'webmock'. But in fact, we can live without any of them, so I remove them.

PR #276, opened 2018-01-04
has_many options (here: class_name)

spape:
_MINOR_

### Options

In case you have to configure relations, the following relation options are available:

`class_name`: Specify the class name of the relation. Use it only if that name can't be inferred from the relation name. So has_many :photos will by default be linked to the Photo class, but if the real class name is e.g. UberallPhoto or namespaced Uberall::Photo, you'll have to specify it with this option.

e.g.

```ruby
module Uberall
  class Location < LHS::Record
    endpoint 'http://uberall/locations'
    endpoint 'http://uberall/locations/:id'
    
    has_many :photos, class_name: 'Uberall::Photo'
  end
end

module Uberall
  class Photo < LHS::Record
  end
end
```


spape:
@Averethel that behaviour for sure, can be added in addition, too. As `class_name` option is also supported by ActiveRecords relation options, I think the "smartness" over configuration is not a blocker here. What do you think?

Averethel:
@spape no it's definitely not a blocker, just something worth considering.

PR #277, opened 2018-01-15
URI templates as in RFC 6750 

Averethel:
_MAJOR_

Updated LHS to version `7` which uses [RFC 6750](https://tools.ietf.org/html/rfc6570) compliant URI template syntax.

This is a breaking change, to update your code you'll need to switch syntax of all defined `endpoint`s from `:<template variable>` to RFC 6750 compliant (usually `{<template variable>}`, if you are using placeholders that you don't want to have urlencoded please use {+<template variable>})`


lhs.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
   s.requirements << 'Ruby >= 2.3.0'
   s.required_ruby_version = '>= 2.3.0'
 
-  s.add_dependency 'lhc', '>= 6.4.0'
+  s.add_dependency 'lhc', '>= 7.0.0.beta1'
```
spape: I guess it's clear, but switch to `>= 7.0.0` before you release ;)

spape:
Also, just like for LHC, add a version description to https://github.com/local-ch/lhs/releases (after release), that
- explains why we switched to the standard
- how to migrate 
👍

PR #293, opened 2018-05-25
Avoid nil data elements

ryabrody:
Error raised when data includes nil elements

# Before
screen shot 2018-05-25 at 18 24 22

# After
screen shot 2018-05-25 at 18 25 04


spape:
BTW: did you try to use LHS 15.2.2, before writing any patch? Looks related: https://github.com/local-ch/lhs/releases/tag/v15.2.2
cc @ralucacril

ryabrody:
Yes, i did try to use 15.2.2 but it did not help as the fix only focuses on the options parameter and not on the data parameter.

PR #307, opened 2018-08-06
Chain error handling integration tests

spape:
This PR adds an integration spec that ensures that error handling in chains actually works as documented.


spec/dummy/app/controllers/error_handling_with_chains_controller.rb
```diff
@@ -0,0 +1,12 @@
+class ErrorHandlingWithChainsController < ApplicationController
+  
+  def handle
+    @records = Record.handle(LHC::Error, ->(error){ error_callback(error) }).find(1)
+  end
+
+  def error_callback(error)
```
ryabrody: should this not be a private method?

spape: true 👍

spape:
@scan @ryabrody 

The initial implementation was not working in Rails 4, so I've ask you to review again.

I had to change how/when the error handler is called in the rails controller, and the following line is crucial to prevent double render errors: https://github.com/local-ch/lhs/pull/307/files#diff-04c6e90faac2675aa89e2176d2eec7d8R824

I also had to refactor the `cleanup` rspec before action a little bit.

PR #313, opened 2018-08-16
Improve retrieving collections from unpaginated endpoints

spape:
## Warn when requesting all from unpaginated endpoints

Fixes: https://github.com/local-ch/lhs/issues/311

When requesting `all` for not paginated endpoints, it's trying to load all pages. If the endpoint is not paginated, LHS loads for ever, as it doesn't know when to stop. Now it shows a warning if it does so, as the behaviour is still desired, for endpoints, that do implement pagination, but do not provide pagination meta data as part of the response.

```ruby
Record.all
# [Warning] "all" has been requested, but endpoint does not provide pagination meta data. If you just want to fetch the first response, use "where" or "fetch".
```

## all with unpaginated endpoints

Fixes: https://github.com/local-ch/lhs/issues/312

In case your record endpoints are not implementing any pagination, configure it to be `paginated: false`. Pagination will not be performed automatically in those cases:

```ruby
# app/models/record.rb

class Record < LHS::Record
  configuration paginated: false
end

```

```ruby

records = Record.all

```
```
  GET https://service.example.com/records
```

## fetch

In case you want to just fetch the records endpoint, without applying any further queries or want to handle pagination, you can simply call `fetch`:

```ruby

records = Record.fetch

```
```
  GET https://service.example.com/records
```

### DIFF
--- a/lib/lhs/concerns/record/chainable.rb
+++ b/lib/lhs/concerns/record/chainable.rb
@@ -16,6 +16,10 @@ class LHS::Record
         Chain.new(self, Parameter.new(hash))
       end
 
+      def fetch
+        Chain.new(self, nil).fetch
+      end
+
       def all(hash = nil)
         chain = Chain.new(self, Parameter.new(hash))
         chain._links.push(Option.new(all: true))
--- a/lib/lhs/concerns/record/request.rb
+++ b/lib/lhs/concerns/record/request.rb
@@ -22,7 +22,7 @@ class LHS::Record
       private
 
       def single_request_load_and_merge_remaining_objects!(data, options, endpoint)
-        return unless options[:all]
+        return if options[:all].blank? || !paginated
         load_and_merge_remaining_objects!(
           data: data,
           options: process_options(options, endpoint),
@@ -40,6 +40,7 @@ class LHS::Record
       # Tries to apply an high value for limit and reacts on the limit
       # returned by the endpoint to make further requests
       def apply_limit!(options)
+        return if !paginated || options[:all].blank?
         options[:params] ||= {}
         options[:params] = options[:params].merge(limit_key(:parameter) => options[:params][limit_key(:parameter)] || LHS::Pagination::Base::DEFAULT_LIMIT)
       end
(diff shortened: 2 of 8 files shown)


spec/record/all_spec.rb
```diff
@@ -101,5 +101,31 @@ def stub_batch(url, items = 10)
       records = Category.all(language: 'en').fetch
       expect(records.length).to eq 300
     end
+
+    it 'outputs a warning, if all is requested, but endpoint does not implement pagination meta data' do
+      stub_batch('http://store/categories?language=en&max=100', 100)
+      stub_batch('http://store/categories?language=en&max=100&offset=100', 100)
+      stub_batch('http://store/categories?language=en&max=100&offset=200', 100)
```
sihu: L:108 could be left away for test-suite optimizations

lib/lhs/concerns/record/chainable.rb
```diff
@@ -16,6 +16,10 @@ def where(hash = nil)
         Chain.new(self, Parameter.new(hash))
       end
 
+      def fetch
+        Chain.new(self, nil).fetch
+      end
```
schmijos: All the other methods return a `Chain` instance. Can it be unexpected for the API user if a method in `Chainable` suddenly resolves?

spape: 'fetch' according to the documentation is about resolving the chain, returning a chain here would be counter intuitive.

schmijos: 👍

lib/lhs/concerns/record/request.rb
```diff
@@ -22,7 +22,7 @@ def request(options)
       private
 
       def single_request_load_and_merge_remaining_objects!(data, options, endpoint)
-        return unless options[:all]
+        return if options[:all].blank? || !paginated
```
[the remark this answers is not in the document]
spape: Than `no_pagination?` would implicitly check if `options[:all]` is set, which, actually has nothing to do with the state of pagination itself and would lead to [coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming))... 

> a method should do one thing and one thing only

At some point we could DRY `options[:all].blank?` to something like `load_additional_data?`.

In the end, I would like to keep the users intend `options`, separated from configuration `paginated` and the pagination state (pagination response meta data).

PR #316, opened 2018-09-20
Made LHS compatible with `form.fields_for` helper

scan:
At one point, `fields_for` uses `to_model`, but first checks with `respond_to?` if the object can handle it. Since `LHS::Data` always returns `true` for any `respond_to?` and then returns `nil` on the actual method, this fails.

So now `to_model` explicitly tells everyone it does not exist.


spape:
and please describe your current scenario/example and what (unwanted) behviour you're facing. thx 🙇

PR #326, opened 2018-12-17
Fix errors information for nested record

zhublik:
This PR fixes the losing of errors in nested records when retrieving associations.


lib/lhs/problems/base.rb
```diff
@@ -87,6 +87,7 @@ def generate_message(attribute, message, _options = {})
 
     def find_translated_message(attribute, message, problem_type)
       normalized_attribute = attribute.to_s.underscore
+      normalized_attribute.slice!(/\d+\./)
```
spape: what's that for?

zhublik: Yeah, i also thought that this is maybe not the best thing to change the `normalized_attribute`, but the purpose of it is to erase numbers from the attribute which we're using to find correct translation.
Example: `appointments.0.date_time`, `appointments.1.date_time`
For this example i need to get it work with the part of I18n key: `appointments.date_time` and in order to do that i have to erase numbers from the `normalized_attribute`

spape: Can you add this in a separate PR with a separate test? Beside that, I don't think that this is necessary, as you should access the objects errors directly... like `appointment_proposals.appointments.first.errors` and then you shouldn't have that problem.

PR #332, opened 2019-01-25
Fix applying nested records endpoint options

spape:
_MAJOR_

Applying nested record endpoint options was not working, because internally, always the options of the root record has been applied:

```ruby
class Account < LHS::Record
  endpoint "https://datastore/accounts"
end

class CustomerRelation < LHS::Record
  endpoint "https://datastore/accounts/{account_id}/customer-relations", auth: { bearer: '123' }
end
```

```ruby
new_account = Account.create(email: 'some@email.com')
```
```
POST https://datastore/accounts
{ customer_relations: { href: 'https://datastore/accounts/123/customer-relations' } }
```

```ruby
customer_relation = new_account.customer_relations
customer_relation.local_customer_id = '123'
customer_relation.save!
```
Was not applying the configured bearer token options on the records endpoint.

# Now

It does apply the records endpoint options as expected.

# Breaking Change
Major version increase, as according to the tests, it might break existing implementations (like booking) that relied on the root record endpoints being used. 

# Migration Path

Add explicit endpoint to the affected nested record. 
See: https://github.com/local-ch/lhs/pull/332/files#diff-45624f27ac5c98e99c577ba56bf33759R90

### DIFF
--- a/lib/lhs/concerns/proxy/accessors.rb
+++ b/lib/lhs/concerns/proxy/accessors.rb
@@ -76,7 +76,8 @@ class LHS::Proxy
     def wrap_return(value, record, name, args = nil)
       name = args.first if name == :[]
       return value unless worth_wrapping?(value)
-      data = (value.is_a?(LHS::Data) || value.is_a?(LHS::Record)) ? value : LHS::Data.new(value, _data, _record, _request)
+      # data = (value.is_a?(LHS::Data) || value.is_a?(LHS::Record)) ? value : LHS::Data.new(value, _data, _record, _request)
+      data = (value.is_a?(LHS::Data) || value.is_a?(LHS::Record)) ? value : LHS::Data.new(value, _data, record, _request)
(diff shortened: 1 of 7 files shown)


lib/lhs/concerns/proxy/accessors.rb
```diff
@@ -76,7 +76,8 @@ def access_collection(value, record, name)
     def wrap_return(value, record, name, args = nil)
       name = args.first if name == :[]
       return value unless worth_wrapping?(value)
-      data = (value.is_a?(LHS::Data) || value.is_a?(LHS::Record)) ? value : LHS::Data.new(value, _data, _record, _request)
+      # data = (value.is_a?(LHS::Data) || value.is_a?(LHS::Record)) ? value : LHS::Data.new(value, _data, _record, _request)
```
zhublik: Did you leave it on purpose or it may be removed?

spape: Forgot about that one 👍 removed it.

PR #333, opened 2019-01-25
Fix class attribute

spape:
The onboarding app was throwing exceptions when this new class_attribute call was executed... 

Requiring active_support core extensions, seem to fix the problem.


scan:
Turns out I cannot deploy until this is released.

PR #335, opened 2019-02-04
Resouce.destroy

ryabrody:
Support `business.destroy` also if business does not have an href.


skaestle:
you may want to explain this behaviour in the README.md

ryabrody:
@skaestle it is already there: https://github.com/local-ch/lhs#destroy-records

PR #345, opened 2019-04-02
Request tracing

spape:
This PR 
- adds support for request tracing for finders: `find find_by find_by! first first! last last!`
Following links does not trigger tracing. 
- allows tracing to be enabled only for `:debug` Rails loglevel

This will be automatically traced:
```ruby
code = AccessCode.find(access_code: params[:access_code]) 
```
```
Called from onboarding/app/controllers/concerns/access_code_concern.rb:11:in `access_code'
```
However, following links will not trigger tracing:
```ruby
code.places
```

```ruby
AccessCode http://storage-stg.preprod-local.ch/customer-onboarding-tokens/FLDRCJBHG
{:href=>
  "http://storage-stg.preprod-local.ch/customer-onboarding-tokens/FLDRCJBHG",
 :token=>"FLDRCJBHG",
 :customer=>
  {:href=>
    "http://lcm-tep.ltvintra.ltv.ch:7002/bsicrmserver/rest/v1/customers/16783561",
   :local_customer_id=>"-FaV6hhzg4UBR6JEfq6ojQ"},
 :places=>
  [{:href=>
     "http://storage-stg.preprod-local.ch/v2/places/evWk_OFY_lF03TVlLqRAow"}],
 :used=>false}
```


lib/lhs/concerns/record/tracing.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class LHS::Record
+
+  module Tracing
+    extend ActiveSupport::Concern
+
+    module ClassMethods
+      # Needs to be called directly from the first method (level) within LHS
+      def trace!(options = {})
+        return options unless Rails.logger.level == 0
```
skaestle: i would like you guys to think about this behaviour being tied to `Rails.logger.level`

If you expect the impact on performance to be non-trivial I think it is even a better argument of having setting, that needs to be set explicitly to enable this functionality. It could even be on request level.

Things to think about:
- what if i want to run my instance on debug but I don't care about LHS traces?
- would you expect a library to depend such a functionality on the LOG level of some other library it happens to use?

lib/lhs/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module LHS
-  VERSION = '19.0.2'
+  VERSION ||= '19.1.0'
```
skaestle: why is this necessary?

skaestle: not sure... maybe this should be a major increase? does the application sufficiently work the same as before in regards to performance for uses that set Rails.logger.level to debug? I would argue that no it is not.

lib/lhs/concerns/record/tracing.rb
technodrome: The idea is that you only need this feature in development and it is not all that noisy. That's why it is bound to `:debug`. When you introduce more flags, switches and knobs, it is just easy to forget to turn off these things. If we see a need to add that in a future based on a real problem, we can always add such functionality. I am just not sure if it is worh solving a problem which we don't even have now and hitting preemptively something that is not even there now.

Logger is a part of activesupport which we use as a dependency IIRC.

lib/lhs/version.rb
technodrome: It's not introducing any breaking changes, just extended functionality. We haven't benchmarked it but on debug it just takes first non-LHS line from caller stack and prints that. I'd be surprised if you could feel any measurable difference while developing. Of course I can't say whether there are any apps running on `:debug` in PROD.

skaestle: I think i came across one or two... but I can't remember for the life of me which ones 
:(

lib/lhs/concerns/record/tracing.rb
skaestle: > The idea is that you only need this feature in development and it is not all that noisy. That's why it is bound to :debug.

I personally think its a fallacy to equal `log_level == DEBUG` to being on an application that is running in a development environment.

skaestle: but I trust you guys with this decision, so this is fine by me ❤️

lib/lhs/version.rb
spape: We are gonna make it an explicit configuration... then we can also stick with minor 👍

technodrome: It's not necessary, just harmonization with LHC.

spec/record/tracing_spec.rb
```diff
@@ -0,0 +1,155 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe LHS::Record do
+
+  context 'tracing' do
+    context 'with level set to debug' do
```
skaestle: one missed

spec/record/tracing_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+        end
+
+        it 'forwards tracing options to lhc' do
+          # binding.pry
+          Place.last
+        end
+      end
+    end
+
+    context 'with level set to other than debug (default in test)' do
```
skaestle: this too

spec/record/tracing_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+
+          # for second reques (last-500)
+          expect(LHC).to receive(:request).with(hash_including(params: { offset: 499, limit: 1 })) do |arguments|
+            expect(arguments).not_to include(:source)
+            spy(:response)
+          end
+        end
+
+        it 'does not forward tracing options to lhc' do
+          Place.last
```
skaestle: this seems to not check anything?

spec/record/tracing_spec.rb
spape: 👍

spec/record/tracing_spec.rb
spape: 👍

spec/record/tracing_spec.rb
spape: expectation was in the before block, moved it to the scenario

PR #350, opened 2019-07-04
Pagination with links

christianarevalo:
To support this kind of pagination with `next` links without offset.

```
{
  "href": "http://storage-stg.preprod-local.ch/v2/localch-accounts?place_id=GXvPAor1ifNfpF0U5PTG0w&limit=10",
  "items": [],
  "limit": 10,
  "previous": {
    "href": "http://storage-stg.preprod-local.ch/v2/localch-accounts?to_user_id=4QE4kSTc-F14x_Sb9B1dMg&place_id=GXvPAor1ifNfpF0U5PTG0w&limit=10"
  },
  "next": {
    "href": "http://storage-stg.preprod-local.ch/v2/localch-accounts?from_user_id=s35N62qOFCW_dtNtd6xuEQ&place_id=GXvPAor1ifNfpF0U5PTG0w&limit=10"
  }
}
```
### Draft implementation & Topics to discuss
- [x] Naming of the new strategy
- [x] Which pagination methods are really required to be implemented? Given that current page, total and counts are not known.
- [x] Using the `next` provided link vs. calculating the link from the received data
- [x] Do we want to support the case when API sends results but doesn't provide a `next` link?

##### Pagination strategy: link

The `link` strategy continuously follows in-response embedded links to following pages until the last page is reached (indicated by no more `next` link).

*WARNING*

Loading all pages from a resource paginated with links only can result in very poor performance, as pages can only be loaded sequentially!

```ruby
# app/models/record.rb

class Search < LHS::Record
  configuration pagination_strategy: 'link'

  endpoint '{+service}/search'
end
```

```ruby

Record.all

```
```
GET https://service.example.com/records?limit=100
{
  items: [{...}, ...],
  limit: 100,
  next: {
    href: 'https://service.example.com/records?from_record_id=p62qM5p0NK_qryO52Ze-eg&limit=100'
  }
}
Sequentially:
  GET https://service.example.com/records?from_record_id=p62qM5p0NK_qryO52Ze-eg&limit=100
  GET https://service.example.com/records?from_record_id=xcaoXBmuMyFFEcFDSgNgDQ&limit=100
```


spape:
A pagination strategy with next and previous should be avoided, as it does not work well with `all` or batch processing in a performant way, as you have to fetch everything sequentially and can't predict and parallelize things beforehand.

Where is this coming from, or how is that triggered?

README.md
```diff
@@ -1217,6 +1217,40 @@ In parallel:
   GET https://service.example.com/records?limit=100&startAt=201
 ```
 
+##### Pagination strategy: link
+
+In comparison to the `offset` strategy, the `link` strategy just increases by 1 (page) and on that page there is a link to the next page and so on until the last page which does not have any link to the next page anyomre.
```
spape: Don't reference other strategies while describing a strategy. Try to only describe `link` here...

e.g.
> The `link` strategy continuously follows in-response embedded links to following pages until the last page is reached (indicated by no more `next` link).

lib/lhs/pagination/link.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  alias count total
+
+  def pages_left
+    pages_left? ? 1 : 0
+  end
+
+  def pages_left?
+    data._raw[:next].present? # TODO use configuration
```
spape: we can make it configurable when we need it to be.

lib/lhs/pagination/base.rb
```diff
@@ -51,6 +55,10 @@ def last_page
       total_pages
     end
 
+    def parallel?
```
spape: We could add a comment

lib/lhs/concerns/record/request.rb
```diff
@@ -277,22 +277,41 @@ def set_nested_data(source, path, value)
       def load_and_merge_paginated_collection!(data, options)
         set_nested_data(data._raw, limit_key(:body), data.length) if data._raw.dig(*limit_key(:body)).blank? && !data.length.zero?
         pagination = data._record.pagination(data)
-        return data if pagination.pages_left.zero?
+        return data unless pagination.pages_left?
         record = data._record
+        if pagination.parallel?
+          load_and_merge_parallel_requests!(record, data, pagination, options)
+        else
+          load_and_merge_sequential_requests!(record, data, options, data._raw.dig(:next, :href), pagination)
```
ryabrody: why not having a `next_href` method in `lib/lhs/pagination/link.rb` for `data._raw.dig(:next, :href)`

ryabrody: why not having a `next_href` method in `lib/lhs/pagination/link.rb` for `data._raw.dig(:next, :href)`

lib/lhs/concerns/record/request.rb
```diff
@@ -277,22 +277,41 @@ def set_nested_data(source, path, value)
...
         end
       end
 
+      def load_and_merge_sequential_requests!(record, data, options, next_link, pagination)
+        warn "[WARNING] You are loading all pages from a resource paginated with links only. As this is performed sequentially, it can result in very poor performance! (https://github.com/local-ch/lhs#pagination-strategy-link)."
+        while next_link.present?
+          page_data = record.request(
+            options.except(:all).merge(url: next_link)
+          )
+          next_link = page_data._raw.dig(:next, :href)
```
ryabrody: why not having a `next_link` method in `lib/lhs/pagination/link.rb` for `data._raw.dig(:next, :href)`

christianarevalo:
Good for me, thank you 👍

lib/lhs/concerns/record/request.rb
spape: this will rather be moved to https://github.com/local-ch/lhs/blame/94cb982fe717d3707fe56709cd8efc5045c8b15a/lib/lhs/concerns/record/configuration.rb#L19 at some point, not to pagination, as we interact with record data here.

Chris and I talked about it, but we are not moving it to configuration, before we are not having at least one example, where next (and previous) would be configured.

PR #351, opened 2019-07-05
Expose Header Information

cedricwider:
### Given
```
    class Location < LHS::Record
      configuration item_key: [:response, :location]
      endpoint 'http://uberall/location'
      endpoint 'http://uberall/location/{id}'
    end
```
A `LHS::Record` which configures a `item_key`.

### Before

We would lose associated information (like the _request object)

### After

We preserve associated information (like the _request object)


spape:
👍

PR #352, opened 2019-07-08
Where with href

spape:
_MINOR_

While `find` was always able to fetch single records by provided href, `where` did not implement that functionality.

This PR adds the ability to pass an href to `where` and fetch a collection of records that way.

This is gonna be used in Customer Center for the user list, where we have to use `next` and `previous` links to navigate the user pagination.

### From the documentation (addition):

You can also provide hrefs to fetch multiple records:

 ```ruby
 Record.where('https://service.example.com/accociation/12345/records')
 ```
```
GET https://service.example.com/accociation/12345/records
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,7 +328,20 @@ Record.where(accociation_id: '12345')
 GET https://service.example.com/accociation/12345/records
 ```
 
-If the provided parameter – `accociation_id` in this case – is part of the endpoint path, it will be injected into the path:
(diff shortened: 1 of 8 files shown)


zhublik:
It's also possible to specify `href` explicitly for `where`? Like `where(href: "https://....")`.

spape:
@zhublik no, as this could overshadow a potential parameter called `href` of a potential future api endpoint.

PR #353, opened 2019-07-18
Option blocks

spape:
Option blocks are a way to apply request options to all the code executed in one block.

## Option Blocks

In order to apply options to all requests performed in a give block, LHS provides option blocks.

```ruby
# app/controllers/application_controller.rb

LHS.options(headers: { 'Tracking-Id' => 123 }) do
  Record.find(1)
end

Record.find(2)
```
```
GET https://records/1 { headers: { 'Tracking-Id' => '123' } }
GET https://records/2 { headers: { } }
```


README.md
```diff
@@ -2315,6 +2317,25 @@ LHS.configure do |config|
   config.request_cycle_cache_enabled = false
 end
 ```
+
+## Option Blocks
+
+In order to apply options to all requests performed in a give block, LHS provides option blocks.
+
+```ruby
+# app/controllers/application_controller.rb
```
ryabrody: I would suggest to use `# app/controllers/records_controller.rb` application_controller is more for stuff which is in general and might confuse here a bit.

lib/lhs/concerns/record/request.rb
```diff
@@ -23,6 +24,11 @@ def request(options)
 
       private
 
+      def deep_merge_with_option_blocks(options)
+        return options if LHS::OptionBlocks::CurrentOptionBlock.options.blank?
+        options.deep_merge(LHS::OptionBlocks::CurrentOptionBlock.options)
```
zhublik: i'm not sure which `deep_merge` is called here, but the one from `active_support` doesn't support `nil` as an argument from her https://github.com/local-ch/lhs/pull/353/files#diff-468569a3307f4bacce8d411d1581c6cbR22 or from `reset_option_block` below.

lib/lhs/concerns/option_blocks.rb
```diff
@@ -0,0 +1,26 @@
...
+      # Will switch to thread_mattr_accessor (which comes with Activesupport) when we dropping support for Active Support v5.
+      extend ActiveSupport::PerThreadRegistry
+      attr_accessor :options
+    end
+
+    module ClassMethods
+      def options(options, &block)
+        CurrentOptionBlock.options = options
+        block.call
+        CurrentOptionBlock.options = nil
```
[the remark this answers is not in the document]
spape: As far as I understand it. The ensure block is only executed if error raising happened. Like rescue first then ensure, but *only* in an exception case. When there is not exception, this line ensures that it gets reset.

lib/lhs/concerns/record/request.rb
spape: But  that's why we return `options` in the line before, when options is blank, to exactly avoid this case. Can you double check?

README.md
spape: good point. improved

lib/lhs/concerns/option_blocks.rb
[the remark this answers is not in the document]
zhublik: No, the code in the `ensure` will always be executed, whether control flows to the rescue block or not.

lib/lhs/concerns/record/request.rb
zhublik: sorry, right 👍

lib/lhs/concerns/option_blocks.rb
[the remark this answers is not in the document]
spape: you're right. I will improve that part

PR #359, opened 2019-08-09
Extended error reporting

spape:
_MINOR_

This PR introduces the `LHS::ExtendedRollbar` interceptor. In combination with a Rollbar processor/handler, it's gonna provide information about all requests/responses prior the exception, when done with LHS.

This PR also refactors how interceptors in LHS are structured in general (was only RequestCycleCache for now).

Screen Shot 2019-08-09 at 4 25 21 PM
Unfortunately, didn't find a way to make this prettier ;)

## Extended Rollbar Logging

In order to log all requests/responses prior to an exception reported by Rollbar in addition to the exception itself, use the `LHS::ExtendedRollbar` interceptor in combination with the rollbar processor/handler:

```ruby
# config/initializers/lhc.rb
LHC.configure do |config|
  config.interceptors = [LHS::ExtendedRollbar]
end
```

```ruby
# config/initializers/rollbar.rb
Rollbar.configure do |config|
  config.before_process << LHS::Interceptors::ExtendedRollbar::Handler.init
end
```


lib/lhs/interceptors/extended_rollbar/thread_registry.rb
```diff
@@ -0,0 +1,18 @@
...
+require 'active_support'
+
+module LHS
+  module Interceptors
+    module ExtendedRollbar
+      extend ActiveSupport::Concern
+
+      class ThreadRegistry
+        # Using ActiveSupports PerThreadRegistry to be able to support Active Support v4.
+        # Will switch to thread_mattr_accessor (which comes with Activesupport) when we dropping support for Active Support v5.
```
zhublik: ...for Active Support v4

lib/lhs/interceptors/request_cycle_cache/interceptor.rb
```diff
@@ -0,0 +1,41 @@
...
+    module RequestCycleCache
+      extend ActiveSupport::Concern
+
+      class Interceptor < LHC::Interceptor
+
+        VERSION = 1
+        CACHED_METHODS = [:get].freeze
+
+        def before_request
+          request.options = request.options.merge({
```
zhublik: you might simplify it with `{ cache: { ... } }.merge(request.options)`

PR #378, opened 2020-03-24
Preload "provider inheriting" records, too

spape:
With the growing usage of [providers](https://github.com/local-ch/lhs#provider) – which reduces endpoint options/configuration redundancy (especially authorization) – the following problem popped up:

LHS needs to preload LHS record classes stored in your application (usually within `models/`) to know all existing endpoints at all time. LHS needs to know all existing endpoints at all time because it needs to infer from a url like `http://storage/places/1` to the corresponding class e.g. `Place` to apply endpoint configurations (like authorization).

The previous auto-loading algorithm was implemented before providers and wasn't loading LHS records which where inheriting from provider classes.

This improves the autoloader to ensure classes inheriting from providers also get preloaded from LHS.

LHS test setup also had to be improved slightly:
- consider providers and multilevel inheritance when reseting LHS cache
- clearly distinguish between LHS records defined in dummy application and those defined in unit tests by adding `Dummy` prefix

### DIFF
--- a/lib/lhs/concerns/autoload_records.rb
+++ b/lib/lhs/concerns/autoload_records.rb
@@ -41,7 +41,7 @@ module AutoloadRecords
           end
 
           Dir.glob(all_model_rb_files).each do |file|
-            next if klasses.none? {|klass|  File.read(file).match(klass)}
+            next if klasses.none? { |klass| File.read(file).match(klass) }
             require_dependency file
           end
         end
--- a/lib/lhs/concerns/autoload_records.rb
+++ b/lib/lhs/concerns/autoload_records.rb
@@ -27,23 +27,27 @@ module AutoloadRecords
           @app.call(env)
         end
 
-        def self.all_model_rb_files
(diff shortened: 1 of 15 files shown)


dimameshcharakou:
One note, which might be related to this (maybe not, though): when i recently setup rails 6 app, then i'm constantly getting the deprecation warning when rails is being initialized: `Rails.application.initialize!`
```
DEPRECATION WARNING: Initialization autoloaded the constants LocalEntry, Chat, and Message.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload LocalEntry, for example,
the expected changes won't be reflected in that stale Class object.

These autoloaded constants have been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
```
And `LocalEntry`, `Chat`, and `Message` are inherited from `LHS::Record`

PR #380, opened 2020-06-15
Automatic OAuth Authentication

spape:
This PR introduces official automatic OAuth authentication.

Adding authentication to all kind of requests made with LHS was the #1 on the annoyance list, when it comes to LHS, often leading people to come up with proxy methods, shared auth options, or storing thread variables themselves, all of it ending up in additional boilerplate code that could've been avoided . Well this PR is gonna take care of it.

## Automatic Authentication (OAuth)

LHS provides a way to have records automatically fetch and use OAuth authentication when performing requests within Rails.

In order to enable automatic oauth authentication, perform the following steps:

1. Make sure LHS is configured to perform `auto_oauth`. Provide a block that, when executed in the controller context, returns a valid access_token/bearer_token.
```ruby
# config/initializers/lhs.rb

LHS.configure do |config|
  config.auto_oauth = -> { access_token }
end
```

2. Opt-in records requiring oauth authentication:

```ruby
# app/models/record.rb

class Record < LHS::Record
  oauth
  # ...
end
```

3. Include the `LHS::OAuth` context into your application controller:

```ruby
# app/controllers/application_controller.rb

class ApplicationController < ActionController::Base
  include LHS::OAuth

  # ...
end
```

4. Make sure you have the `LHC::Auth` interceptor enabled:

```ruby
# config/initializers/lhc.rb

LHC.configure do |config|
  config.interceptors = [LHC::Auth]
end
```

Now you can perform requests based on the record that will be auto authenticated from now on:

```ruby

Record.find(1)
```
```
https://records/1
Authentication: 'Bearer token-12345'
```

### Configure multiple auth providers (even per endpoint)

In case you need to configure multiple auth provider access_tokens within your application,
make sure you provide a proc returning a hash when configuring `auto_oauth`, 
naming every single provider and the responsive method to retrieve the access_tokens in the controller context:

```ruby
# config/initializers/lhs.rb
LHS.configure do |config|
  config.auto_oauth = proc do
    {
      provider1: access_token_provider_1,
      provider2: access_token_provider_2
    }
  end
end
```

Then make sure you either define which provider to use on a record level:

```ruby
# model/record.rb
class Record < LHS::Record
  oauth(:provider1)
  #...
end
```

or on an endpoint level:

```ruby
# model/record.rb
class Record < LHS::Record
  endpoint 'https://service/records', oauth: :provider1
  #...
end
```


README.md
```diff
@@ -2444,6 +2445,66 @@ LHS.configure do |config|
...
+LHS provides a way to have records automatically fetch and use OAuth authentication when performing requests within Rails.
+
+In order to enable automatic oauth authentication, perform the following steps:
+
+1. Make sure LHS is configured to perform `auto_oauth`. Provide a block that, when executed in the controller context, returns a valid access_token/bearer_token.
+```ruby
+# config/initializers/lhs.rb
+
+LHS.configure do |config|
+  config.auto_oauth = -> { access_token }
```
ryabrody, line 2459: Should the access_token not be defined per endpoint? Or how can we set now different access_tokens for different endpoints?

spape: Right now we only do have 1 type of bearer token per user/application pair, as we only have one auth. I think you are talking about client authentication, were we also have e.g. salesforce tokens.

As client_credentials do not need to be configured during runtime (in the controller context, coming from users application session cookie, set by oauth), we do not need automatic oauth for them anyway, as we configure them (still) once per endpoint using local_client_authentication.

Again, this is for OAuth grant type `Authorization Code` not `Client Credentials`, which you can still configure by endpoint:
```
endpoint, '{+resource}/resource/{id}`, auth: { bearer: -> { ClientAuthentication.access_token(refresh: true) } }
```

spape: Once we have multiple auth providers for `Authorization Codes` we can improve this implementation to something like:

```
LHS.configure do |config|
  config.auto_oauth = -> {{ 
    auth_provider_1: access_token_provider_1,
    auth_provider_2: access_token_provider_2
  }}
end

class Record < LHS::Record
  oauth(:auth_provider_2)
end
```

but until we need it, I wouldn't implement it (yet).

spape: and then even by endpoint:

```
endpoint '{+resource}/resource/{id}`, auth: { bearer: -> { LHS::OAuth.access_token(:auth_provider_2) } }
```

spape: added multi auth provider configuration here: https://github.com/local-ch/lhs/pull/380/commits/12f7972a538adf66133b1691584059daea91eadc

allows configuring multiple providers per record or per endpoint.

lib/lhs/concerns/record/configuration.rb
```diff
@@ -13,7 +13,20 @@ module Configuration
...
-        @configuration = args.freeze || {}
+        @configuration = args || {}
+      end
+
+      def auto_oauth?
+        LHS.config.auto_oauth && @configuration && auto_oauth
+      end
+
+      def auto_oauth
+        @configuration.fetch(:auto_oauth, false)
```
dimameshcharakou, line 24: Wouldn't that make sense to check here for `nil` case as well since it's a public method and might be used not only in `auto_oauth?`?

spape: good point. took care of it in here: https://github.com/local-ch/lhs/pull/380/commits/430a505329e5d7d186164aac4ec030fa085cdb89
by also removing all expensive try calls.

PR #392, opened 2020-08-12
compact: Remove included resources that didn't return any records

spape:
Implements `compact` and `compact!` into `LHS::Collection::InternalCollection`.

Also refactors extending data by moving it from bloated `request.rb` to a proper concern of `LHS::Data`.

#### compact: Remove included resources that didn't return any records

In case you include nested data and ignored errors while including, it can happen that you get back a collection that contains data based on response errors:

```ruby

user = User
  .includes(:places)
  .references(places: { ignore: LHC::NotFound })
  .find(123)
```

```
GET http://service/users/123
{ "places": { "href": "http://service/users/123/places" } }

GET http://service/users/123/places
{ "items": [
  { "href": "http://service/places/1" },
  { "href": "http://service/places/2" }
] }

GET http://service/places/1
200 { "name": "Casa Ferlin" }

GET http://service/places/2
404 { "status": 404, "error": "not found" }
```

```ruby
user.places[1] # { "status": 404, "error": "not found" }
```

In order to exclude items from a collection which where not based on successful responses, use `.compact` or `.compact!`:

```ruby

user = User
  .includes(:places)
  .references(places: { ignore: LHC::NotFound })
  .find(123)
places = user.places.compact

places # { "items": [ { "href": "http://service/places/1", "name": "Casa Ferlin" } ] }
```

### DIFF
--- a/lib/lhs/concerns/data/extend.rb
+++ b/lib/lhs/concerns/data/extend.rb
@@ -7,14 +7,14 @@ class LHS::Data
   module Extend
     extend ActiveSupport::Concern
 
-    # Extends already fetched data (self) with additionally 
+    # Extends already fetched data (self) with additionally
     # fetched data (addition) using the given key
     def extend!(addition, key)
-      if self.collection?
+      if collection?
         extend_collection!(addition, key)
       elsif self[key]._raw.is_a? Array
         extend_array!(addition, key)
-      elsif self.item?
+      elsif item?
         extend_item!(addition, key)
       end
     end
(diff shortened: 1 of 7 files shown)


lib/lhs/concerns/data/extend.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
+
+require 'active_support'
+
+class LHS::Data
```
spape, line 5: this just moved here from https://github.com/local-ch/lhs/pull/392/files#diff-ffbc06effcbee79e18771740f1ba521b  with slight adaptions

PR #393, opened 2020-08-17
PRO-1272 Bugfix: Fix Broken Type Mapping

cedricwider:
## Jira

https://jira.localsearch.ch/browse/pro-1272

## Changes

- Add specs to prevent regressions and for faster bugfixing

### DIFF
--- a/lib/lhs/concerns/collection/internal_collection.rb
+++ b/lib/lhs/concerns/collection/internal_collection.rb
@@ -56,13 +56,13 @@ class LHS::Collection < LHS::Proxy
 
       def cast_item(item)
         data = LHS::Data.new(item, @parent, @record)
-        record_by_href(item)&.new(data) || data
+        (record_by_href(item) || @record)&.new(data) || data
       end
 
       def record_by_href(item)
-        return if plain_value?(item) || (item[:href].blank? && @record.blank?)
+        return if plain_value?(item) || item[:href].blank?
 
-        LHS::Record.for_url(item[:href]) || @record
+        LHS::Record.for_url(item[:href])
       end
 
       def plain_value?(item)
--- a/lib/lhs/concerns/collection/internal_collection.rb
+++ b/lib/lhs/concerns/collection/internal_collection.rb
@@ -60,12 +60,12 @@ class LHS::Collection < LHS::Proxy
       end
 
       def record_by_href(item)
-        return if json_value_type?(item) || (item[:href].blank? && @record.blank?)
+        return if plain_value?(item) || (item[:href].blank? && @record.blank?)
 
         LHS::Record.for_url(item[:href]) || @record
       end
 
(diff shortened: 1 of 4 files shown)


spec/record/compact_spec.rb
```diff
@@ -51,17 +66,25 @@ class User < LHS::Record
...
+      expect(places.compact.length).to eq 1
+      expect(places.length).not_to eq 1 # leaves the original intact
+    end
+  end
+
+  context 'record casting' do
+    let(:expected_display_name) { '*Mc Donalds*' }
+
+    it 'finds the right record class' do
+      expect(places.first.display_name).to eq expected_display_name
```
cedricwider, line 79: This passes as you'd expect.

spec/record/compact_spec.rb
```diff
@@ -51,17 +66,25 @@ class User < LHS::Record
...
+      expect(places.map(&:display_name)).to eq [expected_display_name]
```
cedricwider, line 80: ...but this fails for whatever reason. :confused:

lib/lhs/concerns/collection/internal_collection.rb
```diff
@@ -47,18 +47,26 @@ def compact!
...
+      end
+
+      def record_by_href(item)
+        return if json_value_type?(item) || (item[:href].blank? && @record.blank?)
+
+        LHS::Record.for_url(item[:href]) || @record
+      end
+
+      def json_value_type?(item)
+        item.is_a?(String) || item.is_a?(Numeric) || item.is_a?(TrueClass) || item.is_a?(FalseClass)
```
spape, line 69: `json_value_type` seems misleading as neither String, Numeric, True or false have something to do with json (well string does). Would suggest something like `plain_value?`

lib/lhs/concerns/collection/internal_collection.rb
```diff
@@ -47,18 +47,26 @@ def compact!
...
       def cast_item(item)
-        record_by_href = LHS::Record.for_url(item[:href]) if item[:href]
         data = LHS::Data.new(item, @parent, @record)
-        return (record_by_href || @record).new(data) if record_by_href || @record
-        data
+        record_by_href(item)&.new(data) || data
+      end
+
+      def record_by_href(item)
+        return if json_value_type?(item) || (item[:href].blank? && @record.blank?)
```
spape, line 63: there should be no `@record` in record_by_href. as `record_by_href` implies that it provides, well a record by href not by instance.

spape: maybe remove it from `record_by_href` and try something like (in `cast_item`):

```ruby
(record_by_href(item) || @record)&.new(data) || data
```

spape (approved):
gracias

PR #395, opened 2020-08-20
Pass reference not options

spape:
There was wrong passing of `options` instead of `reference` deep inside some of the handle includes internal in cases includes where not pre expanded :atom: 🚀 .

This might accidentally passed `?limit=100` to some included requests, that you need to remove again in your specs.


cedricwider (approved):
Thanks for this fix! :star:

PR #397, opened 2020-10-16
Fix autoloading of inheriting records

dimameshcharakou:
# Description
The issue i faced during the update of `LHS` for `location_app` (some content was stripped on purpose):
```ruby
# app/models/data_layer/favorites.rb

class DataLayer
  class ListType
    def build
      if favorites?
        'Favorites'
      end
    end
  end
end
```
such file will be loaded automatically when there is the following class: `class Favorite < LHS::Record; end`

This PR changes the regexp to more restrict and might match only the whole word(s).

P.S. Not sure how to test this behaviour since the file i put inside `spec/dummy/app/models` will be loaded when tests are executed.

### DIFF
--- a/lib/lhs/version.rb
+++ b/lib/lhs/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module LHS
-  VERSION = '25.0.0'
+  VERSION = '25.1.0'
 end
--- a/lib/lhs/version.rb
+++ b/lib/lhs/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module LHS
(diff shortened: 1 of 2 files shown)


lib/lhs/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module LHS
-  VERSION = '25.0.0'
+  VERSION = '25.1.0'
```
spape, line 4: patch `25.0.1`

dimameshcharakou: ops - thanks 👍

spape (approved):

PR #400, opened 2020-11-18
Expand included links

spape:
When performing includes via multiple levels, .e.g.

```ruby
User.includes(places: :contracts)
```

LHS had problems to continue including nested includes, when there were unexpanded links in between, like places in this example, as `users/id/places` returns just a list of unexpanded place links.

This PR makes sure that LHS expands loaded links before continuing with includes.

### DIFF
--- a/lib/lhs/concerns/collection/internal_collection.rb
+++ b/lib/lhs/concerns/collection/internal_collection.rb
@@ -39,7 +39,7 @@ class LHS::Collection < LHS::Proxy
       def compact
         dup.tap do |collection|
           collection.compact! if collection.raw.present?
-        end
+        end.as_json # do not return an internal collection!
       end
 
       def compact!
--- a/lib/lhs/concerns/data/extend.rb
+++ b/lib/lhs/concerns/data/extend.rb
@@ -9,7 +9,7 @@ class LHS::Data
 
     # Extends already fetched data (self) with additionally
     # fetched data (addition) using the given key
-    def extend!(addition, key=nil)
+    def extend!(addition, key = nil)
       addition = cast_relation_class_for_extension(addition, key)
       if collection?
         extend_collection!(addition, key)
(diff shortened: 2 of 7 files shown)


lib/lhs/concerns/collection/internal_collection.rb
```diff
@@ -39,7 +39,7 @@ def each(&_block)
       def compact
         dup.tap do |collection|
           collection.compact! if collection.raw.present?
-        end
+        end.as_json # do not return an internal collection!
```
spape, line 42: added explicit spec too   
https://github.com/local-ch/lhs/pull/400/files#diff-a53f438f86198623ea15c854a8738ece2c35fa0b625b00a46e63e7bb0e63d761R71

lib/lhs/concerns/data/extend.rb
```diff
@@ -9,11 +9,11 @@ module Extend
 
     # Extends already fetched data (self) with additionally
     # fetched data (addition) using the given key
-    def extend!(addition, key)
+    def extend!(addition, key = nil)
```
spape, line 12: Allows to extend data with data, without extended only a part of it, but rather the entire thing.

lib/lhs/concerns/record/request.rb
```diff
@@ -321,11 +321,22 @@ def load_include_simple!(options, record)
       end
 
       # Continues loading included resources after one complete batch/level has been fetched
-      def continue_including(data, including, referencing)
-        handle_includes(including, data, referencing) if including.present? && data.present?
+      def continue_including(data, included, reference)
+        return data if included.blank? || data.blank?
+        expand_data!(data, included, reference) unless expanded_data?(data)
```
spape, line 326: Before we continue including includes, we now do expand data first, to make sure that an include referencing something that requires expansion first works.

christianarevalo (approved):

PR #406, opened 2021-05-05
Replace deprecated github actions and fix tests.

dimameshcharakou:
# Description
Replace deprecated github actions and fix tests.


spec/dummy/app/controllers/request_cycle_cache_controller.rb
```diff
@@ -19,7 +19,7 @@ def parallel
     render json: users.to_json
   end
 
-  def headers
+  def request_headers
```
dimameshcharakou, line 22: This fixes
```
SystemStackError (stack level too deep):
app/controllers/request_cycle_cache_controller.rb:23:in `headers'
```

ryabrody (approved):

PR #409, opened 2021-09-23
Fix auth options precedence for auto auth

christianarevalo:
_PATCH_

Allows to use auth options which takes precedence over auto auth.
https://github.com/local-ch/lhs#automatic-authentication-oauth

```ruby
Record.options(auth: {. bearer: 'ACCESS_TOKEN'}).find(12345)
```


cedricwider (approved):
:100: YES! Nice. Thanks :pray:

